PR #186, opened 2013-06-17
[RETARGET] Remove Out variable in Response struct and embed http.ResponseWriter

plamb:
This changes the Response struct to embed the http.ResponseWriter. This causes a name conflict with ResponseWriter.WriterHeader and the following error:

```
*Response does not implement http.ResponseWriter (wrong type for WriteHeader method)
        have WriteHeader(int, string)
        want WriteHeader(int)
```

I've renamed the WriteHeader to WriteStatusHeader to avoid the conflict. The other changes are to replace the Response.Out code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,12 +31,17 @@ forked project **will not work**.
 
 Instead, this is the recommended way:
 
-1. Fork the project on Github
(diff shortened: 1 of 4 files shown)


notzippy:
Closing this request as server engine will obsolete it

PR #190, opened 2013-06-19
[RETARGET] Array support by Config.

3d0c:
Here is a simple array implementation. It's possible now to map something like this

```
connection.mysql.drive = mysql
connection.mysql.spec = user:pass@localhost:3306
connection.pgsql.drive = pgsql
connection.pgsql.spec = postgres:pass@dbhost:5432
```

into

```
map[
    mysql:map[
        driver : mysql 
        spec  : user:pass@localhost:3306
    ] 
    pgsql:map[
        driver : postgres 
        spec  : postgres:pass@dbhost:5432
    ]
]
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -111,6 +111,29 @@ func (c *MergedConfig) Options(prefix string) []string {
 	return options
 }
 
+func (c *MergedConfig) Array(prefix string) interface{} {
+	var result interface{} = make(map[string]interface{})
+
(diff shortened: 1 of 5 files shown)


notzippy:
I am looking at a design to implement multiple configuration options, if ini does not suit your needs you will be able to use another, like YAML

PR #245, opened 2013-07-26
Adding in RestfulRoutes

ScruffyProdigy:
CHANGES:
changed parseRoutes() to allow a RESOURCE route method, which will
create a series of other routes

If the Controller has a pluralized name, it will create a route for
each of the following methods it finds in the controller: Index, New,
Create, Show, Edit, Update, Destroy

If the Controller has a singularized name, it will create a route for
these methods instead:
New, Create, Show, Edit, Update, Destroy

QUESTIONS:
Are there any other tests I should write?

COMMENTS:
As most web browsers don't natively have a way to create PUT or DELETE requests, I added the following Filter when I did integration testing right before the RouterFilter:

```
    func(c *revel.Controller, fc []revel.Filter) {
        method := c.Request.Request.FormValue("_method")
        if method != "" {
            c.Request.Request.Method = strings.ToUpper(method)
        }
        fc[0](c,fc[1:])
    }
```

And added `<input type="hidden" name="_method" value="put">` into my HTML forms.  (This is very similar to how Rails performs)

In order to have an out of the box experience, we might need to do something similar
- UPDATE -
  Now uses a lowercase format discussed here: https://groups.google.com/forum/#!topic/revel-framework/MPpD6aJkTvo
  Now uses 'resource' for singular resources and 'resources' for plural
  Fixed the comment styles

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -214,8 +214,7 @@ func (c *Controller) Message(message string, args ...interface{}) (value string)
 func (c *Controller) SetAction(controllerName, methodName string) error {
 
 	// Look up the controller and method types.
-	var ok bool
-	if c.Type, ok = controllers[strings.ToLower(controllerName)]; !ok {
+	if c.Type = GetController(controllerName); c.Type == nil {
 		return errors.New("revel/controller: failed to find controller " + controllerName)
 	}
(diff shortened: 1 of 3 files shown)


pedromorgan:
Hi @ScruffyProdigy, can you target at develop branch so we can consider for merge TIA

PR #271, opened 2013-08-13
[RETARGET] Add {{repeat .content .times}} template tag

MZAWeb:
Adds a _repeat_ template tag that will repeat a given unescaped content a given
number of times. Works like {{pad}} but with a dynamic content.

Use case:

Let's assume you need to show a tree structure, you may want to do something like this:

`{{repeat "<span class='indent'></span>" .Depth}}`

### DIFF
--- a/template.go
+++ b/template.go
@@ -89,7 +89,10 @@ var (
 			}
 			return template.HTML(html.EscapeString(str) + strings.Repeat("&nbsp;", width-len(str)))
 		},
-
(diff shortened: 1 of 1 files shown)


brendensoares:
Sorry for taking forever to come to a decision on this PR.

@revel/core has discussed this and has determined that you can add your own template functions already in Revel and this scenario doesn't appear to be a common need.

Thanks for contributing this idea @MZAWeb!

PR #294, opened 2013-09-05
[RETARGET] Support for 'revel debug' option.

cbonello:
Hi,

I'm new to Revel and quickly realized that having the option to use gdb would speed up my learning process.

I'm able to start gdb with my patch but I can't see the original source code while debugging. I tried to run the go compiler with -gcflags "-N -l" (http://golang.org/doc/gdb) but it does not help. 

I'm aware of https://github.com/robfig/revel/issues/33 but could not figure out what to do with '-srcPath'.

Any idea of what I should do next?

Thanks,

Ch. Bonello

### DIFF
--- a/revel/debug.go
+++ b/revel/debug.go
@@ -1,0 +1,90 @@
+package main
+
+import (
+	"fmt"
+	"github.com/robfig/revel"
+	"github.com/robfig/revel/harness"
+	"os"
+	"os/exec"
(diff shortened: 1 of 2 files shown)


notzippy:
There is documentation on running revel in intellij which can provide some debugging. The best debugging is good logging. Closing - thanks

PR #305, opened 2013-09-12
Support for different character tags in the templates

marcinsnt:
Supports various characters tags in the templates according to TemplateLoader.

Example:

mail.html:

``` html
[[define "show"]]
ok
[[end]]
[[template "show"]]
```

invoice.html:

``` html
<<define "invoice">>
ok
<<end>>
<<template "invoice">>
```

app.go:

``` go
func (c App) Mail() r.Result {
    tl := revel.NewTemplateLoader([]string{"/project/name/views/mail"}, "/project/name/views/mail", "[[ ]]")
    tl.Refresh()
    tpl, err := tl.Template("mail.html")
    if err != nil {
        return c.RenderError(err)
    }
    return &revel.RenderTemplateResult{
        Template:   tpl,
        RenderArgs: c.RenderArgs,
    }
}

func (c App) Invoice() r.Result {
    tl := revel.NewTemplateLoader([]string{"/project/name/views/invoice"}, "/project/name/views/invoice", "<< >>")
    tl.Refresh()
    tpl, err := tl.Template("invoice.html")
    if err != nil {
        return c.RenderError(err)
    }
    return &revel.RenderTemplateResult{
        Template:   tpl,
        RenderArgs: c.RenderArgs,
    }
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -55,7 +55,7 @@ func Run(port int) {
 		address = fmt.Sprintf("%s:%d", address, port)
 	}
 
-	MainTemplateLoader = NewTemplateLoader(TemplatePaths)
+	MainTemplateLoader = NewTemplateLoader(TemplatePaths, ViewsPath, "")
 
 	// The "watch" config variable can turn on and off all watching.
 	// (As a convenient way to control it all together.)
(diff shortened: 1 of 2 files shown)


notzippy:
I get this, but since we broke out templating to use template engines, we can now provide a different way of addressing this issue on even a per file basis. Closing this

PR #306, opened 2013-09-13
1. Update skeleton to Bootstrap 3.0.0; 2. Add new html file into views t..

haibinzhang:
1. Update skeleton to Bootstrap 3.0.0  and update related html's CSS.
2. Update jQuery to 1.10.2
3. Add nav.html, sidebar.html and skeleton.html into app/views directory, to minimize html repetition by including skeleton.html. 
4. Reformat all html to generate neat html source code.


brendensoares:
We'll update bootstrap and jquery in a new PR and target the correct branch.

PR #358, opened 2013-10-24
[RETARGET] supports subdomain

xiaoao:
I think this isn't a graceful way to support subdomain, and  who will be nice enough to give me a better way
thx in advance


brendensoares:
See also: #536

notzippy:
Closing this one in favor of hosts

Issue #554, opened 2014-03-14
Pluggable logger


pushrax:
For production applications, having structured log output is a must (e.g. JSON). It would be cool if you could set up Revel to send its logs out to a formatter before printing.

Since there is no standard "Logger" interface, implementing this is a bit annoying. A co-worker wrote https://github.com/Sirupsen/logrus which I really like, but it wouldn't be great to introduce a dependency just for this. One option could be to copy out this interface:

``` go
type StandardLogger interface {
    Print(...interface{})
    Printf(string, ...interface{})
    Printfln(...interface{})

    Fatal(...interface{})
    Fatalf(string, ...interface{})
    Fatalln(...interface{})

    Panic(...interface{})
    Panicf(string, ...interface{})
    Panicln(...interface{})
}
```

and set it to a `log.Logger` by default.

Alternatively, we can introduce our own formatter interface, with the default being Revel's current formatting.

Thoughts?

janekolszak:
Maybe this will help anyone: https://github.com/janekolszak/revfluent
It's a small Revel + Fluentd logger.

littlexiang:
seems writing a module to override the revel.LOGGER is a good way

PR #557, opened 2014-03-16
Make getLogger public and allow it to be overridden

pushrax:
Addresses #554. Here's how I'm using it:

``` go
type logrusInfoWriter struct {
    target *logrus.Logger
}
func (l *logrusInfoWriter) Write(p []byte) (n int, err error) {
    l.target.Info(strings.TrimSpace(string(p)))
    return len(p), nil
}

// etc.

func getLogger(name string) *log.Logger {
    switch name {
    case "info":
        // Hijack stream to point to a different logger.
        return log.New(&logrusInfoWriter{Log}, "", 0)
    case "warn":
        // etc.
    }

    panic("Unknown logger type: " + name)
}

func init() {
    revel.GetLogger = getLogger
}
```

Total hack, but covers my needs completely. Feel free to discuss.

### DIFF
--- a/revel.go
+++ b/revel.go
@@ -187,10 +187,10 @@ func Init(mode, importPath, srcPath string) {
 	}
 
 	// Configure logging.
-	TRACE = getLogger("trace")
-	INFO = getLogger("info")
-	WARN = getLogger("warn")
-	ERROR = getLogger("error")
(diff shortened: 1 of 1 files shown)


notzippy:
#1213 replaces this.

Issue #553, opened 2014-03-14
Expose http.Handler


bobobo1618:
Sorry if this has already come up before, I feel like I've wanted it long enough that everyone should have argued to their heart's content.

Anyhow, I'd like for Revel to expose the `http.Handler` and allow another package to set up the `main` method and run the application.

Exposing it in this manner would open up a lot of possibilities, particular with integration with other libraries and the Go ecosystem in general (AppEngine is among the things that could be done with this I believe).

Plus, it lets other people take care of things like HTTPS (#549) by just running Go's built in `ListenAndServeTLS`.

Without this, I feel that Revel's cutting itself out of a large portion of the ecosystem, distancing itself needlessly from the great work being done by other projects.

notzippy:
Issue #1156 allows for a custom server engine which will allow any implementation

PR #648, opened 2014-06-24
Feature/expose revel http handler

brendensoares:
Progress toward completing revel/revel#553
- [ ] fix single line imports in skeleton
- [ ] update sample apps to build with new structure
- [ ] decide how to initialize the `revel.Server` without running Revel


notzippy:
The revel project has been split up since this PR, #1156 issue will allow the user to plug in their own template engine

Issue #261, opened 2013-08-07
Support for zero downtime upgrades


redalastor:
It should use or be somewhat like goagain:

https://github.com/rcrowley/goagain

You start your new process and the old one dies when it finished serving all the pending requests (and doing all the pending jobs in Revel's case).

notzippy:
The server engine is abstracted now, so this may now be customized using a custom server engine implementation

PR #690, opened 2014-08-16
Support zero downtime restarts

pushrax:
Getting the ball rolling for #261. Seems to work well for me, but I haven't tested it in production yet. As with unix sockets, SSL is not (yet) supported.

@xthexder


paritosh90:
@brendensoares Have we come up with any solution on this.
@pushrax The main file is generated dynamically at compile time in tmp. Can we change it without breaking process?

PR #728, opened 2014-09-17
Add default http.Status code for render methods.

dmitrykakunin:
We need check `response.Status` in filter chain, but the current implementation doesn't set status except `NotFound`, `Forbidden` and `Todo` methods. I have added default statuses for other render methods.


brendensoares:
Looking at it once again, I have decided to merge this PR as it is a step in the right direction. We can improve the code as we go, but we must make steady progress.

brendensoares:
Thanks @dmitrykakunin!

PR #746, opened 2014-10-03
Add i18n templates.

klauspost:
All templates can now have an optional language suffix. The loader will attempt to load a template with the language specified language as an extension to the specified name. For instance "index.html" with "en" as language will attempt to load "index.html.en" first, and fall back to "template.html" if the file cannot be found.

Internationalized templates can also be dynamically included in the templates. To do this, use a the template function {{ i18ntemplate . "included.html" }}

This function will also attempt to load "included.html.en" if the language is English, and will fall back to including "included.html" if the language cannot be found.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -103,9 +103,12 @@ func (c *Controller) Render(extraRenderArgs ...interface{}) Result {
 // A less magical way to render a template.
 // Renders the given template, using the current RenderArgs.
 func (c *Controller) RenderTemplate(templatePath string) Result {
+	// Get Language
+	// We ignore if cast is unsuccessful, since it will give us an empty string anyway.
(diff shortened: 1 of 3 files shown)


brendensoares:
After talking this over with @jeevatkm we feel it would be best to not merge this PR and instead include this functionality in the upcoming pluggable template engine.

Thank you @klauspost for contributing this solution, it's a valid idea that we can use soon!

klauspost:
@pedromorgan ... you are closing because.....?

pedromorgan:
See @brendensoares comment above and pluggable template engine

notzippy:
Adding functionality to PR #1170, thanks for your input and design.

PR #768, opened 2014-10-21
i18n: Make Message Translation plugable

klauspost:
This allows users to override the Message function in revel with their
own. It defaults to the existing Message function, so there should be no
compatibility issues.


klauspost:
The use case was replacing i18n with an auto-reloading one as well as setting locale through url path.

brendensoares:
I see some potential improvements to `i18n.go`, like renaming `messages` to `messageConfigs` and making a `messages map[string]map[string]string` set to the parsed values.

For now, this PR is backwards compatible, so it looks great!

Thanks @klauspost!

PR #772, opened 2014-10-24
Feature/graceful shutdown

brendensoares:
Retarget of #593. Not currently functioning due to missing `StopServer()` method.


alkchr:
Standard `net/http` will have graceful shutdowns built-in starting from Go 1.8.

zuoRambo:
so, the latest version (master branch) still don't support `graceful shutdown`?

notzippy:
Closing , will need to redo the code to implement this since the age of this PR is so old

PR #800, opened 2014-11-12
Add an option to trim whitespace from rendered HTML.

klauspost:
Add an option for trimming the generated HTML, which will do the following:
 * Remove all leading & trailing whitespace on every line
 * Remove all empty lines
 * Attempt to keep formatting inside < pre > < /pre > tags

This is fairly safe unless "white-space: pre;" is used in css for formatting. Since there is no way to detect that, you will have to keep trimming off in these cases.

### DIFF
--- a/results.go
+++ b/results.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"code.google.com/p/go.net/websocket"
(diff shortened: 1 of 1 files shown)


brendensoares:
I think, to your point @klauspost, we can make this modular, like @anonx is wanting, during v0.13 which would then allow us to better unit test the code. I think for now, this code looks good to go.

Thanks @klauspost!

PR #809, opened 2014-11-23
Static module: get prefix from fixed params explicitly

ghost:
We have discussed it with @pushrax. I'll add additional details to this PR a bit later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 A high productivity, full-stack web framework for the [Go language](http://www.golang.org).
 
-Current Version: 0.11.1 (Oct 27, 2014)
+Current Version: 0.11.2 (Nov 23, 2014)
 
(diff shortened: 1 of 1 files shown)


brendensoares:
I've been out of the loop for a bit too long. I'm not sure what the issue is here, though I trust it's a real issue.

Could I have some steps to reproduce this along with the expected results versus the actual results? That way I can better contribute to the conversation.

I have to catch up on a lot of other issues, so I'll watch for this thread. Thanks.

anonx:
@brendensoares ``http://localhost:9000/Static/Serve?prefix=/etc&filepath=passwd`` or ``prefix=conf&filepath=app.conf``. This worked on any project that had a wildcard route.

brendensoares:
@anonx I read through the related issues and now I see what's happening. Hope most people are using nginx or the like :disappointed:

PR #824, opened 2014-12-07
Fix for static module files

nadeemc:
Getting the prefix from fixed parameters was overriding the absPath and
causing 404 on static files for modules (e.g. testrunner)


brendensoares:
@anonx no reason we can't do another patch as long as it's not a breaking change. That's the point of patch versioning :)

PR #828, opened 2014-12-15
Search modules for message files as well

notzippy:
This patch searches the modules for message files as well. This enhances modules so they can also provide translations with the packages.

### DIFF
--- a/i18n.go
+++ b/i18n.go
@@ -91,6 +91,14 @@ func loadMessages(path string) {
 	if error := filepath.Walk(path, loadMessageFile); error != nil && !os.IsNotExist(error) {
 		ERROR.Println("Error reading messages files:", error)
 	}
+
+	// Read in messages from the modules
+	for _, module := range Modules {
+		TRACE.Println("Importing messages from ", filepath.Join(module.Path, messageFilesDirectory))
+		if error := filepath.Walk(filepath.Join(module.Path, messageFilesDirectory), loadMessageFile); error != nil && !os.IsNotExist(error) {
+			ERROR.Println("Error reading messages files:", error)
+		}
+	}
 }
 
 // Load a single message file


brendensoares:
@notzippy can you please provide steps to verify that this works from a new app to the expected results?

Please & thank you.

notzippy:
Steps, I create a module called `bar` and add a `messages/test.en` file to it to translate `helloworld`. Note this just adds `<b>A friendly {{msg . "helloworld"}}</b>` to the created view in the application, in the real world messages in the module should be normally only used in that module.

	revel new temp/foo
	mkdir -p $GOPATH/src/temp/foo/modules/bar/messages
	echo "helloworld=Hi there"> $GOPATH/src/temp/foo/modules/bar/messages/test.en
	echo "module.bar=temp/foo/modules/bar" >>$GOPATH/src/temp/foo/conf/app.conf
	echo "<b>A friendly {{msg . \"helloworld\"}}</b>" >> $GOPATH/src/temp/foo/app/views/App/Index.html 
	revel run temp/foo prod

(Had to run in production mode since the module is appended to the production section of the app.conf)

Additional notes, 
1) It is expected that the developer follow the same guidelines with the messages as they would with the views, ie prefix all the messages with something appropriate which would not conflict with another message since these are appended to the "global" message queue

anonx:
So, here is a short summary:
**Revel** can import from modules (http://revel.github.io/manual/modules.html): 
1. ``templates``
2. ``controllers``
3. ``routes``
4. ``assets`` (using ``Static.ServeModule`` method).
This PR makes it possible to import ``i18n`` files, too. @notzippy Is my understanding correct?

cast @pedromorgan

notzippy:
Yes this allows modules to declare there own 18n messages. I was not aware
that routes could be defined in a module, how does one so that?

anonx:
@notzippy https://github.com/revel/revel/blob/master/samples/booking/conf/routes#L5 and #422.

brendensoares:
@anonx @notzippy importing module routes is in [the Revel manual](http://revel.github.io/manual/routing.html#routing-modules) under "Routing Modules".

notzippy:
@brendensoares @anonx thx, learn something new every day...

i18n.go
```diff
@@ -91,6 +91,14 @@ func loadMessages(path string) {
 	if error := filepath.Walk(path, loadMessageFile); error != nil && !os.IsNotExist(error) {
 		ERROR.Println("Error reading messages files:", error)
 	}
+
+	// Read in messages from the modules
+	for _, module := range Modules {
```
brendensoares: It may be better to add a `MessagesPaths` similar to `TemplatePaths` in `revel.go` instead of accessing Modules directly. Thoughts? It's a minor difference, but more consistent with the current module convention.

notzippy: Messages is a different beast to begin with. It is loaded through an `init` method and `OnAppStart`. Templates are loaded in server.go:run. To me neither seem totally correct, exposing TemplatePaths seems unnecessary (as does ConfPaths and CodePaths). My thoughts are that the revel:addModule method should have a callBack list of items to be notified when a module is added, (with the possibility of adding more callbacks in the config). Each item then would take care of initializing its own area. However I do not want to even look at this until after we have broken completed the breaking up of revel. 

For the intermediate time though I think this is a valuable PR even if it is not implemented in the same way as the templates.

pedromorgan:
@brendensoares Can u merge the develop branch into master ASAP, people are referinging stuff.. and we need to go!!!!!!!!!!!!

or shall i commit to master myself.. he thinks !!!

brendensoares:
@pedromorgan what do you mean "referinging stuff"?

Don't merge into `master`, that would break our workflow. The Roadmap for v0.12 has been reduced, we just need to finish it up and then we can merge into `master` and release `develop`. There isn't much effort needed before we can release v0.12.

notzippy:
Will this make the 0.12 release ? Or will it be deferred till later ?

brendensoares:
@notzippy No, this PR is not the correct way to go about loading i18n files in modules.

We can still use this PR, but we need to make some changes.

# TODO
- [ ] Track a `MessagesPath` in `revel.go`'s `func addModule(name, importPath, modulePath string)` that `i18n.go` can search for messages files.

I'm rescheduling this for the next release as we don't have time to update this PR and reverify it. v0.12 needs to get buttoned up and out the door.

anonx:
@brendensoares We are using Modules directly in `static` module BTW: https://github.com/revel/revel/blob/master/modules/static/app/controllers/static.go#L66
The only difference is here messages are scanned once (on bootstrap time). And `static` is iterating over modules on every request which is not really cool if we have many modules.

notzippy:
@brendensoares Why is it important to store the messagesPath ? After the message dictionary is built is there any need for the path to be stored ? Perhaps if I understood the usefulness of this it would make more sense.

brendensoares:
@anonx @notzippy I'm thinking that we have a user configurable `MessagesPath` that is searched for message files. We may also have a single user configurable flag, such as `SearchModulesForMessages`, that would then adds all detected modules' path to `MessagesPath`.

"Separation of Concerns" seems like a good pattern to follow in general and this is an opportunity to use it. Agreed?

brendensoares:
Thanks @notzippy and @jeevatkm for feedback.

Issue #837, opened 2015-01-03
Too many open files - connections do not close


opennota:
I run out of open files limit (1024 files), because the connections seem to not close. A part of `lsof -n -p PID` output follows:

    opennota 16458 opennota   71u  IPv6 43159075      0t0    TCP [myserver]:9000->[client-ip]:53836 (ESTABLISHED)
    opennota 16458 opennota   72u  IPv6 43158889      0t0    TCP [myserver]:9000->[client-ip]:50512 (ESTABLISHED)
    opennota 16458 opennota   73u  IPv6 43155244      0t0    TCP [myserver]:9000->[client-ip]:64589 (ESTABLISHED)
    opennota 16458 opennota   74u  IPv6 43158891      0t0    TCP [myserver]:9000->[client-ip]:50520 (ESTABLISHED)
    opennota 16458 opennota   75u  IPv6 43159077      0t0    TCP [myserver]:9000->[client-ip]:53838 (ESTABLISHED)
    opennota 16458 opennota   76u  IPv6 43155330      0t0    TCP [myserver]:9000->[client-ip]:50257 (ESTABLISHED)
    opennota 16458 opennota   77u  IPv6 43155332      0t0    TCP [myserver]:9000->[client-ip]:50258 (ESTABLISHED)
    opennota 16458 opennota   78u  IPv6 43155334      0t0    TCP [myserver]:9000->[client-ip]:50259 (ESTABLISHED)
    opennota 16458 opennota   79u  IPv6 43158893      0t0    TCP [myserver]:9000->[client-ip]:50523 (ESTABLISHED)
    opennota 16458 opennota   80u  IPv6 43159079      0t0    TCP [myserver]:9000->[client-ip]:53837 (ESTABLISHED)
    opennota 16458 opennota   81u  IPv6 43158895      0t0    TCP [myserver]:9000->[client-ip]:50522 (ESTABLISHED)
    opennota 16458 opennota   82u  IPv6 43159081      0t0    TCP [myserver]:9000->[client-ip]:53835 (ESTABLISHED)
    opennota 16458 opennota   83u  IPv6 43159083      0t0    TCP [myserver]:9000->[client-ip]:53834 (ESTABLISHED)
    opennota 16458 opennota   84u  IPv6 43171822      0t0    TCP [myserver]:9000->[client-ip]:59842 (ESTABLISHED)
[...]

PR #843, opened 2015-01-07
Feature/sane http timeout defaults

brendensoares:
This is in response to the conversation on #837. Also, I snuck in a little, unrelated comment for `flash.go`.

### DIFF
--- a/flash.go
+++ b/flash.go
@@ -12,6 +12,7 @@ import (
 // E.g. the Post/Redirect/Get pattern:
 // http://en.wikipedia.org/wiki/Post/Redirect/Get
 type Flash struct {
+	// `Data` is the input which is read in `restoreFlash`, `Out` is the output which is set in a FLASH cookie at the end of the `FlashFilter()`
 	Data, Out map[string]string
 }
 
--- a/server.go
+++ b/server.go
@@ -94,8 +94,10 @@ func Run(port int) {
 	}
 
 	Server = &http.Server{
-		Addr:    localAddress,
-		Handler: http.HandlerFunc(handle),
+		Addr:         localAddress,
+		Handler:      http.HandlerFunc(handle),
+		ReadTimeout:  time.Minute,
+		WriteTimeout: time.Minute,
 	}
 
 	runStartupHooks()


anonx:
@brendensoares was that fix related to [extending of timeouts on new requests rather than on connections](https://code.google.com/p/go/source/diff?spec=svn6f26fec4b5cd3facf3d2a16b131ac37ae7aad1f9&name=6f26fec4b5cd&r=6f26fec4b5cd3facf3d2a16b131ac37ae7aad1f9&format=side&path=/src/pkg/net/http/server.go) you've mentioned [here](https://github.com/revel/revel/issues/837#issuecomment-68800334) ever applied to golang?

I'm trying to test chat sample (websockets) using revel (develop) and go1.4:
```sh
revel run github.com/revel/samples/chat
```
And it looks like it works not as we expect it. I'm actively sending new messages, but the connection is closed exactly 1 minute after openning the websocket connection but not 1 minute after the last message.

You can try it yourself. Start the chat sample. Using Firebug console or something add the following line and keep sending some messages.
```javascript
socket.onclose = function() { alert("Closed!"); }
```
The connection will be closed (and you will get "Closed!" alert) despite the fact you're actively using the chat.

brendensoares:
@anonx That [`net/http` issue](https://code.google.com/p/go/issues/detail?id=4676) I mentioned was committed almost 2 years ago it seems. Why do you ask?

If it's not working how we hoped, then what do we do? Is it possible it does work as expected, but not for web socket connections?

brendensoares:
@anonx are you connecting directly to Revel or via a proxy? Are you binding to localhost or an IP? I wonder if this is an issue with the WS library we're using? Can you verify what your Go runtime's `src/pkg/net/http/server.go` file contains on line 419 versus [the diff we're looking at](https://code.google.com/p/go/source/diff?spec=svn6f26fec4b5cd3facf3d2a16b131ac37ae7aad1f9&name=6f26fec4b5cd&r=6f26fec4b5cd3facf3d2a16b131ac37ae7aad1f9&format=side&path=/src/pkg/net/http/server.go)?

anonx:
@brendensoares You are right. Everything works fine but the websocket part.

Here we are creating our initial handshake HTTP connection.
```js
var socket = new WebSocket('/app/websocket')
```
*Here is the code responsible for its processing:* https://github.com/revel/revel/blob/master/server.go#L25

The connection above is the only HTTP connection which is used by websocket chat. There're no any subsequent HTTP requests. We just initializing our websocket and it does all the stuff for us.
That's why the fix doesn't affect websocket part. And no matter whether you're  actively using websocket or not, the initial connection will be timed out. No HTTP requests -> no timeout extension.

Websocket doesn't do anything with our initial HTTP connection's timeout (doesn't extend on subsequent websocket requests). Don't know whether it is by design or it's a bug. Have to read the specs to find out.

brendensoares:
I have confirmed that the read/write timeout on `revel.Server` does affect websocket connections, too.

I made the following changes to `server.go`:
```
func handle(w http.ResponseWriter, r *http.Request) {
	upgrade := r.Header.Get("Upgrade")
	if upgrade == "websocket" || upgrade == "Websocket" {
		websocket.Handler(func(ws *websocket.Conn) {
			ws.SetDeadline(time.Now().Add(time.Second * 10))
			var msg = make([]byte, 512)
			if _, err := ws.Read(msg); err != nil {
				ERROR.Println(err)
			}
			WARN.Println(time.Now(), string(msg))
			r.Method = "WS"
			handleInternal(w, r, ws)
		}).ServeHTTP(w, r)
	} else {
		handleInternal(w, r, nil)
	}
}
```
I noticed that `WARN.Println()` is only called once on the initial websocket frame. The key was to set the deadline of the websocket connection via `ws.SetDeadline(time.Now().Add(time.Second * 10))` which sets it to 10 seconds into the future. I tried longer intervals into the future with great success.

Ideally, we want to extend the deadline for each websocket frame, [just like the `net/http` server does](http://golang.org/src/net/http/server.go?s=47492:48985#L575), but I'm not sure the best way to accomplish this. It seems the [websocket `Codec`](http://godoc.org/golang.org/x/net/websocket#Codec) may be the proper place to do this.

I've created a PR to collaborate on this solution: #860.

brendensoares:
Solution noted in #860. This PR is done done.

PR #853, opened 2015-01-14
[RETARGET] Fixed router app start bug without routes.

sevki:
App doesn't crash when it doesn't have any routes to serve, and the
returned error is discarded, not checked. This checks if the error is
not nil and crashes if it should because no routes means nothing to do.


brendensoares:
@sevki thanks for contributing to Revel!

Please retarget your PR to the `develop` branch per [our contribution guidelines](https://github.com/revel/revel/blob/master/CONTRIBUTING.md#how-revel-is-developed).

After the PR is targeting the correct branch, I would like to know the steps to reproduce this issue so we can confirm what problem is being solved and that it is working as expected.

Thank you!

sevki:
Sorry about that. Will do.

PR #855, opened 2015-01-16
Fixes router app start bug without routes.

sevki:
App doesn't crash when it doesn't have any routes to serve, and the
returned error is discarded, not checked.This checks if the error is
not nil and crashes if it should because no routes means nothing to
do.

### DIFF
--- a/router.go
+++ b/router.go
@@ -425,8 +425,9 @@ func init() {
 		MainRouter = NewRouter(path.Join(BasePath, "conf", "routes"))
 		if MainWatcher != nil && Config.BoolDefault("watch.routes", true) {
 			MainWatcher.Listen(MainRouter, MainRouter.path)
-		} else {
-			MainRouter.Refresh()
+		} else if err := MainRouter.Refresh(); err != nil {
+			// Not in dev mode and Route loading failed, we should crash.
+			ERROR.Panicln(err.Error())
 		}
 	})
 }


sevki:
Steps to reproduce:
What turned out to be the problem was, I was running in a mode in where `mode.dev` was enabled, testrunner module was in the routes but the testrunner thing was not in the conf file itself. 

This arrangement worked fine if I did `revel run app mymode` but didn't work when I did `app -runMode mymode -importPath app` all I got were [No matching route found: /](https://github.com/sevki/revel/blob/master/router.go#L439) errors.

Upon more investigation I found out that the `MainRoute.Routes` were empty, in my case because it couldn't find the testrunnermodule in the coressponding section in the conf, but this could have been also due to a corrupt routes file -- which mine technically was -- due to some other error like a missing comma, illegal char or no like breaks, etc. 

If not for anything this fixes the unchecked error which in my case the addition helped debug an error.

brendensoares:
@sevki thanks for contributing this and sharing the details of the issue. This will be included in v0.12 once the scheduled roadmap items are completed.

anonx:
Tests [fail with the error](https://travis-ci.org/revel/revel/builds/47267385):

> --- FAIL: TestBenchmarkCompressed (0.01 seconds)
>
> panic: Route validation error (in /home/travis/gopath/src/github.com/revel/modules/jobs/conf/routes:1): revel/controller: failed to find controller Jobs
>
> [recovered]
>
> panic: Route validation error (in /home/travis/gopath/src/github.com/revel/modules/jobs/conf/routes:1): revel/controller: failed to find controller Jobs

Looks like [TestBenchmarkCompressed](https://github.com/revel/revel/blob/master/compress_test.go#L10) has to be updated. Does [fake booking app](https://github.com/revel/revel/blob/master/fakeapp_test.go#L61) require modifications?
Oh, and did I tell you I don't like how revel is tested?

anonx:
OK, here is the problem with tests. This PR adds check of routes to `OnAppStart` func: https://github.com/revel/revel/blob/3fb3e8973c461b41e8a6575a1b2468560b2e6598/router.go#L424

And the fake app starts `OnAppStart` funcs first and only after that it initializes routes: https://github.com/revel/revel/blob/master/fakeapp_test.go#L70
```go
// Fake app.
 func startFakeBookingApp() {
	...
	// Line 70: (STARTUP HOOKS ARE HERE)
	runStartupHooks()

	// Line 72:
	MainRouter .... blah-blah-blah
	routesFile ... blah-blah-blah
	...
}

// Our startup hook.
func ThisPR() {
	if routerIsEmpty {
		ERROR.Fatal("Go away!")
	}
}
```

The order of initializations and stuff like that in fake app should be identical to the order we have in real apps:
https://github.com/revel/revel/blob/master/server.go#L58

anonx:
OK, looks like we cannot just move `runStartupHooks()` (line 70) below all those router related stuff. Because one of the startup hooks checks whether loaded routes are correct. And not all of them are correct as our fake app implements only a few controllers out of many real booking sample has to. So, we would get the error:

> Route validation error (in /path/to/go/src/github.com/revel/modules/jobs/conf/routes:1): revel/controller: failed to find controller Jobs

brendensoares:
:+1: good work @anonx thanks!

Issue #418, opened 2013-12-20
track current Revel version


brendensoares:
Along with git tagging, we should add a `VERSION` file in the root of the revel package and expose it to Revel apps.

pedromorgan:
Also #474

pedromorgan:
Having it in a version.txt file file solved a lot of porblem with flightgear... compiling docs, cmake et all...

In the meantime, one possibility for the docs is to use a data file..
like
https://github.com/pedromorgan/revel.github.io/blob/pedro/_data/version.yaml
and appears on the homepage
http://revel-docs.daffodil.uk.com/

brendensoares:
@pedromorgan I really like that YAML file idea. Is that a jekyll feature? Can it read from remote domains?

pedromorgan:
yes the _data directory is a jekyll feature..
http://jekyllrb.com/docs/datafiles/

Dont think it will work on remote domain, but could be scripted I suppose

brendensoares:
The PR#858 may not be perfect, but it's present. Closing.

PR #858, opened 2015-01-19
Adds first draft of a version file to Revel

brendensoares:
Parent: #418 

## TODO
- [x] add `version` file
- [x] add runtime parsing of version file and output to console

### DIFF
--- a/revel.go
+++ b/revel.go
@@ -201,6 +201,7 @@ func Init(mode, importPath, srcPath string) {
 	loadModules()
 
 	Initialized = true
+	INFO.Printf("Initialized Revel v%s (%s) for %s", VERSION, BUILD_DATE, MINIMUM_GO)
 }
 
 // Create a logger using log.* directives in app.conf plus the current settings
--- a/version.go
+++ b/version.go
@@ -1,0 +1,10 @@
+package revel
+
+const (
+	// Current Revel version
+	VERSION    = "0.12.0dev"
+	// Latest commit date
+	BUILD_DATE = "2015-01-19"
+	// Minimum required Go version
+	MINIMUM_GO = ">= go1.3"
+)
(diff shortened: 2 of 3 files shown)


pedromorgan:
Ok OK.. cool,, but now I realise a mistake, probably my fault..

Snag is that the remote "docs compiler" and jenkins etc,, need the actual source and name..

So what would work, after thinking about it..
```
name: "Revel Core" << name of package
source: github.com/revel/revel  <<  location
version: 0.1.2 << version no
date: 2015-12-24T11.59.59Z << date of version as panic release..
```

I know its just but more complicated.. but my python
fab get_revel failed.. as we need more meta..

Incidentally modules would be
```
name: "Auth Module" << name of package
source: github.com/revel/modules/foo  <<  location
version: 0.1.1 << version no
date: 2015-11-24T11.59.59Z << date of version as panic release..
```

anyway.. having it in Yaml is fantastic.. or in welsh 
```
"bendigedig" = rejoice and happy

ben di ge dig
head dizzy and high
```

brendensoares:
I'd say this is now ready to merge. We can add a `revel version` command soon. @anonx @pushrax any feedback?

brendensoares:
@revel/core I may be losing my mind, but is it even possible to consume the `version.yaml` file that we added? @pedromorgan is this consumed by the website or docs? Why did we add this? We should add a comment to this file as far as what it's purpose is.

alkchr:
@brendensoares IIRC it is not consumed. We discussed it in IRC about a year ago. The plan was to start using it.

pedromorgan:
The yaml is part of the documentation, and a way to determine what the "code is". Eg info on a new module or alike added to the documentation

brendensoares:
@pedromorgan ok, could you provide more info for how we can make use of it? For example, the yaml file is not included in the `revel package` command, should it be?

PR #860, opened 2015-01-22
Adds debug code to reproduce websocket connection timeout issue

brendensoares:
Bug introduced by #843. See its thread for more details.

## TODO
- [x] ~~Ideally, we want to extend the deadline for each websocket frame, [just like the `net/http` server does](http://golang.org/src/net/http/server.go?s=47492:48985#L575), but I'm not sure the best way to accomplish this. It seems the [websocket `Codec`](http://godoc.org/golang.org/x/net/websocket#Codec) may be the proper place to do this.~~

The app can extend the deadline for each message received without Revel's help. A sane deadline of 24 hours is set by default for WS requests now.


anonx:
:+1:

PR #865, opened 2015-01-31
Map to index action of controllers when no action is specified.

carc1n0gen:
I figured other people would want the ability to route to index action of controller by default when none is provided.

### DIFF
--- a/skeleton/conf/routes
+++ b/skeleton/conf/routes
@@ -12,5 +12,8 @@ GET     /favicon.ico                            404
 # Map static resources from the /app/public folder to the /public path
 GET     /public/*filepath                       Static.Serve("public")
 
+# Map to index action of controllers by default when no action is specified
+GET     /:controller/                           :controller.index
+
 # Catch all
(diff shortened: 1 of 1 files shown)


carc1n0gen:
Sorry meant to add a description.  I figured other people would want the ability to route to index action of controller by default when none is provided.

anonx:
@carc1n0gen Will it show 404 or 500 if `index` method does not exist?

carc1n0gen:
It displays 500 with the message "Template post/index.html not found."

pushrax:
I don't think either of these are a good idea:

```
GET     /:controller/                           :controller.index
*       /:controller/:action                    :controller.:action
```

The potential security downside is pretty high.

brendensoares:
I tend to agree with @pushrax on this issue. I don't much like catch all routes like this. I would remove the current catch all instead of adding another.

brendensoares:
@carc1n0gen thanks for taking the time to share this with the community. I'm closing since it's an easy change for a dev to make themselves.

PR #867, opened 2015-02-05
Add a symlink-aware filepath.Walk helper

### LINKED ISSUE #673: Symbolic Walker
Per @pushrax and @verdverm's comments on #555, we need a general walker function that is able to walk symbolic links. This is used in multiple places noted below (copied from #555):

draft: https://gist.github.com/verdverm/75b036439591ac0ef47e

we should probably call it:   `revelWalk`

There are walk functions, which make use of `filepath.Walk`, in...
 - `watcher.go`
 - `template.go`
 - `i18n.go`
 - `harness/reflect.go`

metafex:
This adds a symlink-aware filepath.Walk helper to unify all the code which needs this functionality in revel.
It fixes #673 and an according pull request for revel/cmd is here: https://github.com/revel/cmd/pull/20.

I tested it with a new app, symlinked the whole folder into another directory (my original use-case) and symlinked some static files around. Everything worked as expected, revel still found all the files and the app built and ran fine.

### DIFF
--- a/util.go
+++ b/util.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strings"
@@ -169,4 +170,30 @@ func Equal(a, b interface{}) bool {
 		}
 	}
 	return false
+}
+
+// Walk is a filepath.Walk helper which follows symlinks and always returns the
+// path of the file or directory that is referenced in the link.
+func Walk(root string, walkFn filepath.WalkFunc) error {
+	var symWalkFn = func(path string, info os.FileInfo, err error) error {
+		var isLink bool
+		// get the target if it is a symlink
+		if err == nil && info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			isLink = true
+			path, err = filepath.EvalSymlinks(path)
+			if err != nil { // abort if we can't evaluate the symlink (e.g. a dangling symlink)
+				return err
+			}
+			info, err = os.Lstat(path)
+		}
+		if err := walkFn(path, info, err); err != nil {
+			return err // abort on walkFn error too (same behavior as filepath.Walk)
+		}
+		// recurse into symlinked directories
+		if err == nil && info.IsDir() && isLink {
+			return Walk(path, walkFn)
+		}
+		return nil
+	}
+	return filepath.Walk(root, symWalkFn)
 }
--- a/i18n.go
+++ b/i18n.go
@@ -2,11 +2,12 @@ package revel
 
 import (
 	"fmt"
-	"github.com/robfig/config"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/robfig/config"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


brendensoares:
Thanks for this @metafex! I want to verify your testing just to be careful, but this looks good at first glance.

pedromorgan:
Can we create test for this  and merge ?

brendensoares:
@pedromorgan anyone can create a test to verify this. Anyone who has time that is. Hopefully it'll be done soon.

util.go
```diff
@@ -170,3 +171,29 @@ func Equal(a, b interface{}) bool {
 	}
 	return false
 }
+
+// Walk is a filepath.Walk helper which follows symlinks and always returns the
+// path of the file or directory that is referenced in the link.
+func Walk(root string, walkFn filepath.WalkFunc) error {
```
pavanka: I found this patch while searching for symlink aware dir traversal. Thank you.
There are a few issues with this: I implemented an alternative code here: https://gist.github.com/pavanka/e22a432fd6487fc8e578

util.go
```diff
@@ -170,3 +171,29 @@ func Equal(a, b interface{}) bool {
...
+		// get the target if it is a symlink
+		if err == nil && info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			isLink = true
+			path, err = filepath.EvalSymlinks(path)
+			if err != nil { // abort if we can't evaluate the symlink (e.g. a dangling symlink)
+				return err
+			}
+			info, err = os.Lstat(path)
+		}
+		if err := walkFn(path, info, err); err != nil {
```
pavanka: so walkFn is called twice on a symlinked directory.

util.go
```diff
@@ -170,3 +171,29 @@ func Equal(a, b interface{}) bool {
...
+
+// Walk is a filepath.Walk helper which follows symlinks and always returns the
+// path of the file or directory that is referenced in the link.
+func Walk(root string, walkFn filepath.WalkFunc) error {
+	var symWalkFn = func(path string, info os.FileInfo, err error) error {
+		var isLink bool
+		// get the target if it is a symlink
+		if err == nil && info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			isLink = true
+			path, err = filepath.EvalSymlinks(path)
```
pavanka: consider a directory structure like 

test
├── a
│   ├── hello
│   └── hi
├── b -> a
└── hi -> a/hi

ideally the names of files one gives to walkFn should be:

test
test/a
test/a/hello
test/a/hi
test/b
test/b/hello
test/b/hi
test/hi

but this code always calls walkFn with fully resolved paths. making it less useful in the general case

util.go
```diff
@@ -170,3 +171,29 @@ func Equal(a, b interface{}) bool {
...
+				return err
+			}
+			info, err = os.Lstat(path)
+		}
+		if err := walkFn(path, info, err); err != nil {
+			return err // abort on walkFn error too (same behavior as filepath.Walk)
+		}
+		// recurse into symlinked directories
+		if err == nil && info.IsDir() && isLink {
+			return Walk(path, walkFn)
```
pavanka: this code cannot detect loops in the directory structure -> so if loops do exist it runs the risk of going into infinite recursion.

for that I wrote some loop detection code & another alternative implementation for the whole code. [it mostly copies from filepath.Walk]

https://gist.github.com/pavanka/2cd9f21639446d53b7b6

pavanka:
for beginners -> if one does not want to detect loops, etc.
https://github.com/pavanka/symwalk/

Issue #614, opened 2014-05-20
Organizing files into sub-packages


brendensoares:
@revel/core What do you think about organizing the go files in the root directory into sub-packages?

I tried simply moving them into sub-folders, but in order for them to share the `revel` package they have to be in the same folder (this is odd and slightly irritating).

We should be able to do this without breaking compatibility and without introducing circular dependencies. Doing this should increase our separation of concerns and hopefully improve code quality through clear lines of division and unit testing.

I also feel that reorganizing into sub-packages will make contribution to Revel more inviting to the Go community and/or new comers.

pedromorgan:
Where are we now with this..

I think there are inbuilt packages/modules eg jobs, testrunner
and then third party

?
its a config issue ??
http://revel-docs.daffodil.uk.com/manual/modules.html

and a manual for others..  

maybe we need `revel create mod` to do that ?

brendensoares:
@pedromorgan see the [Roadmap for v0.12](https://github.com/revel/revel/wiki/Roadmap#v012-see-issues) and specifically the "Create sub-packages" checkbox.

I would like to create a few more internal packages, but @anonx brings up some good points that I haven't thought through fully.

@anonx my high level goal is to have a clear separation of concerns. I'd like to do this without affecting the exported interface that Revel devers see. I want it to be a purely internal change. I will admit that sub-packages may not be the correct solution for each of the items listed above and I am ready to find more reasonable solutions if we can think of them.

As far as criteria for creating a sub-package, I'd say if a segment of code is independently functional and complex enough (say ~300 lines of code or more) then we should seriously consider its re-organization. The criteria is somewhat subjective though.

PS - I am inclined to split v0.12 into 2 releases as it's too large currently. It's more important that we keep consistent momentum than trying to release large improvements. Agreed?

brendensoares:
@anonx it just struck me that we may not be understanding sub-packages the same way. Not all that I've suggested on the v0.12 Roadmap should be external packages. Internal, independently testable packages are more appropriate than external for most, if not all, of the remaining items. Agreed?

anonx:
@brendensoares I'm asking about the criteria just for practical reasons. We have ``init.go`` with the following content:
```go
revel.Filters = []revel.Filter{
	...
	revel.RouterFilter,
	revel.SessionFilter,
	...
}
```
I, as a user of the framework, expect that every standard filter will have predictable import path. For example:
```go
import (
	"github.com/revel/revel/routing"
[...]

PR #868, opened 2015-02-09
Extracting config into its own subpackage

alkout:
#614 is a related discussion.

#### This PR includes:
* `config.go` has been moved to its own `config` package
* Things have been renamed:
  * `MergedConfig` to `config.Context`
  * `NewEmptyConfig` to `config.NewContext`
  * `LoadConfig` to `config.LoadContext`

Moreover, `.travis.yml` has been updated (`go test revel...`) so every subpackage will be tested automagically.

Looks like, `revel/cmd` *doesn't* need any changes. So, ~~need a revel/cmd issue~~.


config/config.go
```diff
@@ -56,18 +57,17 @@ func (c *MergedConfig) Int(option string) (result int, found bool) {
 	}
 
 	// If it wasn't an OptionError, it must have failed to parse.
-	ERROR.Println("Failed to parse config option", option, "as int:", err)
 	return 0, false
 }
 
-func (c *MergedConfig) IntDefault(option string, dfault int) int {
+func (c *Context) IntDefault(option string, dfault int) int {
```
anonx: `dfault` sounds like something bad :) I'll replace it by `default`.

brendensoares:
:+1: this looks great @anonx! Didn't even think about TravisCI config. Good catch.

@revel/core Thoughts?

notzippy:
Slightly unrelated @anonx, how come all your notifications are "From: Fuck
You <notifications@github.com>"

pedromorgan:
Can we just merge this to "dodgy" and see what happens ??

brendensoares:
@pedromorgan there is no `dodgy`. Everything we merge should be confirmed as working and ideally have automated unit tests.

anonx:
Who's going to test this? This PR is all about renaming. Nothing new is introduced.

brendensoares:
@anonx in this case we'll need manual peer testing to confirm all works as expected and that there are no regression bugs. That's all.

config/config.go
```diff
@@ -1,31 +1,32 @@
...
 
-func LoadConfig(confName string) (*MergedConfig, error) {
+func LoadContext(confName string, confPaths []string) (*Context, error) {
 	var err error
-	for _, confPath := range ConfPaths {
+	for _, confPath := range confPaths {
 		conf, err := config.ReadDefault(path.Join(confPath, confName))
 		if err == nil {
-			return &MergedConfig{conf, ""}, nil
+			return &Context{conf, ""}, nil
```
notzippy: Since we are modifying LoadConfig should we not pass in the section as well instead of defaulting it to "" ? 
Also small peeve of mine is initialization of structures without field names `return &Context{conf,""}` imho should be `return &Context{config:conf}`

config/config.go
```diff
@@ -34,19 +35,19 @@ func LoadConfig(confName string) (*MergedConfig, error) {
 	return nil, err
```
notzippy: Somehow this `for` loop appears wrong as well, if an `err` is assigned something by config.ReadDefault it is ignored the variable `err` is created inside the for loop so it is impossible to reach line 32. I know you didnt change this part but we may as well correct its behavior.

    for _, confPath := range confPaths {
        if conf,err := config.ReadDefault(path.Join(confPath, confName)); err!=nil && err!="not found" {
              return err
        } else if err==nil {
               return &Context{config:conf}, nil
        }
    }
    return nil, fmt.Errorf("Config %s not found",confName)

config/config.go
anonx: > Since we are modifying LoadConfig should we not pass in the section as well instead of defaulting it to "" ? 

@notzippy Doesn't matter. Lack of section string is [interpreted](https://github.com/robfig/config/blob/master/type.go#L106) as `DEFAULT` anyway.

> Also small peeve of mine is initialization of structures without field names return &Context{conf,""} imho should be return &Context{config:conf}

Ideally we should also `go lint` the code, cover by comments, and add more tests. But to my view of thinking we should keep PRs simple (one change = one PR). Though I'm not always following this rule.

config/config.go
anonx: @notzippy You are right. But I think we can just remove `var err error` declaration and `if err == nil` check.
```diff
func LoadContext(confName string, confPaths []string) (*Context, error) {
-	var err error
	for _, confPath := range confPaths {
		conf, err := config.ReadDefault(path.Join(confPath, confName))
		if err == nil {
+			// Line 29:
			return &Context{conf, ""}, nil
		}
	}
-	if err == nil {
-		err = errors.New("not found")
-	}
+	It is guaranteed err is not nil here. Otherwise, we would exit at line 29.
	return nil, err
}
```

For, reference here is an initial version of this function: https://github.com/revel/revel/blob/43d05c83b5464e64329f10aba010351a4fe44156/config.go#L16-L22
The change was introduced by https://github.com/revel/revel/commit/86302320e2b9fe4257f352112ab337b3e2276cd0.

notzippy: If you remove `var err error` this will not compile since the creation of `err` variable is only within the scope of the for loop.

anonx: @notzippy Right, here is the final version:
```go
func LoadContext(confName string, confPaths []string) (*Context, error) {
 	var err error
	var config config.Config
	for _, confPath := range confPaths {
		conf, err = config.ReadDefault(path.Join(confPath, confName))
		if err == nil {
			return &Context{conf, ""}, nil
		}
	}
	return nil, err
}
```

notzippy: :+1:

anonx:
I assume all interested parties have tested it and we are ready to merge?

alkout:
I'm removing all my repos so clone [config_package](https://github.com/anonx/revel/tree/config_package) branch and probably [cachesession](https://github.com/anonx/cachesession) / [cacheflash](https://github.com/anonx/cacheflash) repos if you're interested. Otherwise, this will become unmergable and the repos will pass into oblivion.

Issue #869, opened 2015-02-10
Router.Reverse wildcard action


notzippy:
In the reverse of the wildcard action defined in the router file I was having an issue. Rather then substituting a string in place of the wildcard it would add an new parameter. Consider the following

Template
`{{url "Foo.Bar" "Hithere"}}`

Routes
`* /foo/*something App.Bar`

Controller
`func (a *App) Bar(something string) revel.Result {...}`

The url is reversed to
`/foo/*something?something=Hithere`

Thoughts ?

notzippy:
@brendensoares Looks like there was a slight oversight in the router (see patch). Not sure if you are moving the router to another spot, if you are can you apply the patch as well ? thx

PR #870, opened 2015-02-10
Wildcard reverse routing parameter patch

notzippy:
This resolves issue #869 . Revel incorrectly would append a parameter to a route path with a wild card (like /foo/*bar) in it.

### DIFF
--- a/router.go
+++ b/router.go
@@ -375,7 +375,7 @@ func (router *Router) Reverse(action string, argValues map[string]string) *Actio
 			pathElements = strings.Split(route.Path, "/")
 		)
 		for i, el := range pathElements {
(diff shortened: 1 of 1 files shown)


pushrax:
Can you add a regression test for this one?

notzippy:
So you would like a test added to  `router_test.go` ? How is that test file launched ? 
`revel run github.com/revel/revel ????`

anonx:
@notzippy, no that is a regular golang test file. So, it's `go test github.com/revel/revel/...`.

brendensoares:
@notzippy @pushrax @pushrax I added a test to cover this issue and merged it as #870.

PR #871, opened 2015-02-11
Feature/binding subpackage

brendensoares:
## TODO
- [x] add time formats to `params` package from `revel` vars that gets passed on to `binding` package
- [ ] convert log messages to returned error result and checking so that revel can log messages
- [ ] solve `Params` dependency
- [ ] update docs https://revel.github.io/manual/binding.html


brendensoares:
Wrong branch (of all people to do this!)

PR #872, opened 2015-02-11
Creates binding subpackage

brendensoares:
## TODO
- [x] add time formats to `params` package from `revel` vars that gets passed on to `binding` package
- [ ] convert log messages to returned error result and checking so that revel can log messages
- [ ] solve `Params` dependency
  - [x] add binder driver clean up method to `Binder` interface to allow deallocation of resources (e.g. tmpFiles (which will be extracted from the `Params` type to free up `binding`'s dependency)
- [ ] update docs https://revel.github.io/manual/binding.html
- [ ] add missing default time formats to `binder_test.go`


brendensoares:
I am increasingly concerned about the naming of this package. `binding` is not concerned with ["data binding"](https://en.wikipedia.org/wiki/Data_binding) in the common definition of that phrase (2-way data binding to keep UI up to date with underlying data models and their attributes).

It is, as [documented in the manual](https://revel.github.io/manual/binding.html), concerned with converting Go types to and from strings.

It seems the most familiar naming to adopt would be `marshalling` as that is what [data conversion is called](https://en.wikipedia.org/wiki/Marshalling_%28computer_science%29).

Assuming we can all agree that this is the best thing to do, we can leave `revel.Params.Bind()` as is for now so that we don't break the exported Revel API (yet). This seems practical since the code subject to this PR is not exported (`func binderForType()`) for use by app code.

Thoughts?

notzippy:
I would be okay with the change to marshal

params.go
```diff
@@ -37,6 +40,7 @@ func ParseParams(params *Params, req *Request) {
 	case "application/x-www-form-urlencoded":
 		// Typical form.
 		if err := req.ParseForm(); err != nil {
+			// TODO: should this be a warning or a panic?
 			WARN.Println("Error parsing request body:", err)
```
anonx: > // TODO: should this be a warning or a panic?

[ParseForm](http://golang.org/src/net/http/request.go?s=23467:23502#L769) may return one of the following types of errors:
- `missing form body`
- `http: POST too large`
Moreover, `ioutil.ReadAll()` and `url.ParseQuery()` which are used by `ParseForm` may return their errors:
- `ioutil.ReadAll`: If the buffer overflows, we will get `bytes.ErrTooLarge`
- `ParseQuery`: EscapeError

I think Revel should return 400 Bad Request if it's not able to parse a request. If that's not possible here, I think it's better to panic (as long as it is "recovered") rather than going on silently as if nothing happened.

revel.go
```diff
@@ -16,6 +16,8 @@ import (
 
 const (
 	REVEL_IMPORT_PATH = "github.com/revel/revel"
+	DEFAULT_DATE_FORMAT     = "2006-01-02"
+	DEFAULT_DATETIME_FORMAT = "2006-01-02 15:04"
```
anonx: @brendensoares Have you decided to restore old time format (#878)? Then we should make changes to config's template as well (https://github.com/revel/revel/blob/master/skeleton/conf/app.conf.template#L69). Because that's kind of counter intuitive to have different built in and config defaults.

anonx:
> convert log messages to returned error result and checking so that revel can log messages

Can we do this with filters? They do not return anything. They import revel for it's logger and they define something like `Params` or `Session` which is used by revel. So, we end up having something like:
```go
// Something type **is used** by revel package.
type Something struct {
}

// SomeFilter is a filter that **uses** revel package.
func SomeFilter(c *revel.Controller, fc []revel.Filter) {
	...
	revel.WARN.Println("Some message here...")
	...
}
```
And thus we've got an import cycle.

pedromorgan:
Can u add
```
DEFAULT_TIME_FORMAT     = "15:04"
```
and that would be exactly what I use already..

notzippy:
I think this is a good idea, but with the code base changes it is best to redo this work - sorry @brendensoares

Issue #733, opened 2014-09-19
Raise up the browser' download dialog?


xpbliss:
In app.conf ,   i set "results.compressed=true" .
In controller,  return a "nil" revel.Result
then on browser, visit the controller page,  it always raise up a  download dialog. it is bothered in InterceptFun or InterceptMethod.

PR #873, opened 2015-02-11
Change compression handling

birkirb:
There are HTTP clients that go bananas when seeing a compression header on an empty response body.
In order to increase compatibility with these, I've made the compression conditional on the status code.

I'd like to add a test for this as well, but didn't see any 204, 304 in the fake apps. If you could give me some instructions I'd be happy to try and follow up on that.


anonx:
> There are HTTP clients that go bananas when seeing a compression header on an empty response body.

@birkirb, what exactly do they do?

birkirb:
In our case we had a problem with Samsung Galaxy Ace, running Android v2.3.6 (Gingerbread). Seems like their HTTP library would try to unzip a null (empty body) or something like that and throw an exception. I saw some anecdotal evidence suggesting it was also on issue on other systems but don't have any details.  

Since the body is already empty it seems like it shouldn't hurt to skip the compression.

anonx:
@birkirb I wonder, wasn't it something related to this https://github.com/revel/revel/issues/733#issuecomment-56713379?

birkirb:
Seems similar, although in that case it's that the controller action there is returning nil and not explicitly setting the content type etc. I've set the status and content type explicitly before returning in the cases where I use this.  HTTP standard stipulates there MUST not be a body on 204 and 304s, so again there is little point in trying to compress it.

brendensoares:
@birkirb we can get this merged for v0.13, but first you need to target the `develop` branch.

birkirb:
@brendensoares New request opened in #888, please close when ready.

PR #877, opened 2015-02-17
Configurable max body request size

JaTochNietDan:
Hi guys,

This is actually a critical feature that we need as we will sometimes have requests coming into us which are larger than the default 10 MB allowed size.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1,7 +1,6 @@
 package revel
 
 import (
-	"code.google.com/p/go.net/websocket"
 	"fmt"
 	"io"
 	"net"
@@ -9,6 +8,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"code.google.com/p/go.net/websocket"
 )
 
 var (
@@ -21,6 +22,7 @@ var (
 // This method handles all requests.  It dispatches to handleInternal after
 // handling / adapting websocket connections.
 func handle(w http.ResponseWriter, r *http.Request) {
+	r.Body = http.MaxBytesReader(w, r.Body, int64(Config.IntDefault("http.maxRequestSize", 10000000)))
 	upgrade := r.Header.Get("Upgrade")
 	if upgrade == "websocket" || upgrade == "Websocket" {
 		websocket.Handler(func(ws *websocket.Conn) {


anonx:
Hello, @JaTochNietDan. I wonder what problem this PR solves?

anonx:
> This is actually a critical feature that we need as we will sometimes have requests coming into us which are larger than the default 10 MB allowed size.

Hello, @JaTochNietDan. I wonder, what is the exact use case? And what is the behaviour of Revel without this patch?

JaTochNietDan:
Essentially we are accepting emails being sent to us via POST request for one of our products (I work at [Teamwork](http://www.teamwork.com)) and we use Revel for parts of our product's backend. The issue is that sometimes emails can grow quite large (generally caused as you continue to reply to emails, the size of the email body grows exponentially as each reply retains the previous thread of replies).

In this case we were unable to process some of these requests as the max body size of a request in Go's http server is 10 MB and one of the emails coming in was 11.4 MB (see [here](http://golang.org/pkg/net/http/#Request.ParseForm)). When this function fails to parse the form in Revel, it prints out a warning and then it seems that we can't access the values in the form (see [here](params.go#L39)).

There may be a better solution for this that I'm not aware of, though this has solved our problem for now.

ready4god2513:
@anonx Is there an ETA on getting this one pulled in?  Or a reason why it shouldn't be?

anonx:
Hi, @ready4god2513. I don't know. I'm a member of Revel Framework's Community just like you are. Let's ask @brendensoares or @pushrax.

ready4god2513:
Ah, my bad @anonx

pushrax:
Merged in https://github.com/revel/revel/commit/bcce705fe25b4a0bc36c13ce3c864ce9f08e8d41, and made a couple modifications in https://github.com/revel/revel/commit/fcf88855b6ec0bb3393a641f84a45610c70edb29. I switched `http.maxRequestSize` to `http.maxrequestsize` to align with the lowercase convention of other settings, and set the default to "don't set this at all."

ready4god2513:
Thanks!

brendensoares:
The only delay is getting `develop` merged into `master`. I'm trying to get the roadmap as complete as possible before releasing v0.12 (especially the new subpackage roadmap items).

Issue #549, opened 2014-03-08
Remove HTTPS from Revel


brendensoares:
As mentioned by @pushrax, Revel is currently handling functionality that it doesn't need to, such as HTTPS.

We need to consider what Revel could do without to enhance separation of concerns and increase the focus of Revel.

@revel/core Thoughts?

brendensoares:
https://github.com/revel/revel/issues/918#issuecomment-89431129 @jniltonho wants to keep HTTPS support.

notzippy:
Server engine allows this to be done

PR #879, opened 2015-02-23
Add InitServer function returning the http handler

metafex:
This is a minimalist approach to somewhat solve #648, #553 and #549. It is minimally invasive and just moves around some code.

I'm looking for feedback on how this could be improved or generally if that's an approach worthy of merging :-)

An example of how this can be integrated into an app lives here: https://github.com/metafex/revel-modules/tree/master/server. Personally I'm running this in production now and don't have any problems with it yet.

The example server module depends on https://github.com/revel/cmd/pull/22 to generate the necessary code in the main function.

### DIFF
--- a/server.go
+++ b/server.go
@@ -59,6 +59,33 @@ func handleInternal(w http.ResponseWriter, r *http.Request, ws *websocket.Conn) 
 	}
 }
 
+// InitServer intializes the server and returns the handler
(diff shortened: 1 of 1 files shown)


brendensoares:
@metafex I'm very curious to look at this in more detail! I'll try to reply tomorrow after work, but first, sleep. :)

brendensoares:
:+1: great PR

Issue #729, opened 2014-09-17
Adding define inside the template results in an error


NitroKKX:
When adding a code like `{{define "T1"}}ONE{{end}}` results in an immediate error `template: redefinition of template` this is because of the code in revel/template.go:287 

```
        lowerCaseTemplateName := strings.ToLower(templateName)

        err = addTemplate(templateName)
        err = addTemplate(lowerCaseTemplateName)
```

Every template is added twice to the set, once as lower and once as upper. This should be changed. I propose to only add the the template as declared in code (and not add the case insensitive version). Then modify the TemplateLoader:Temaplate function to the following

func (loader *TemplateLoader) Template(name string) (Template, error) {
    // Look up and return the template.
    tmpl := loader.templateSet.Lookup(name)
    if tmpl == nil {
        // Lower case the file name to support case-insensitive matching
        tmpl = loader.templateSet.Lookup(strings.ToLower(name))
    }

This seems to work well for my purposes. The case insensitive match may be redundant.

PR #885, opened 2015-02-27
Optimize search of template names

anonx:
This PR addresses https://github.com/revel/revel/pull/753#commitcomment-9943539 and #729.

Behaviour of `Template` method from #753 should not be affected.
* If we have `index.html` and `Index.html` only one of them will be registered as template. And the second one will be ignored.
* If we have `{define "test"}` and `{define "TEST"}` the first declaration will be used and the second one ignored.
* Both `c.RenderTemplate("app/index.html")` and `c.RenderTemplate('App/Index.html')` work no matter what register `app` dir and `index.html` file have.

The only change is a map of lower case names is generated once. The map is then used to find a real name of template (whatever register it has). Thus, number of iterations to find the template we need (on average) will be decreased.

### DIFF
--- a/template.go
+++ b/template.go
@@ -29,6 +29,8 @@ type TemplateLoader struct {
 	paths []string
 	// Map from template name to the path from whence it was loaded.
 	templatePaths map[string]string
+	// templateNames is a map from lower case template name to the real template name.
+	templateNames map[string]string
 }
 
 type Template interface {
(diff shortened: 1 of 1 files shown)


brendensoares:
:+1 this looks thorough, but deserves more focused consideration.

anonx:
v0.13? C'mon, the change is super easy, just a few lines but important. Let's double check now and include in v0.12.

brendensoares:
@anonx I hope we're not going to run into similar issues we've had in the past with this topic (#745 #395)

PR #888, opened 2015-03-04
Change compression handling

birkirb:
See #873

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -44,6 +44,7 @@ type CompressResponseWriter struct {
 }
 
 func CompressFilter(c *Controller, fc []Filter) {
+	fc[0](c, fc[1:])
 	if Config.BoolDefault("results.compressed", false) {
 		if c.Response.Status != http.StatusNoContent && c.Response.Status != http.StatusNotModified {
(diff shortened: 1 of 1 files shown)


brendensoares:
I want to get this included in v0.12, which I'm trying to button up now and release.

However, I don't like that this overrides the filter ordering. I feel strongly that this should be [controlled by the filter ordering](https://github.com/revel/revel/blob/develop/skeleton/app/init.go#L18) as defined by the developer. What is your motivation with this change?

notzippy:
@brendensoares I would think the reasoning behind forcing this to the last is so that the entire response is compressed, ie no extra data added afterwards. Otherwise the filter is pretty much useless because if you add something after it, the compression is lost.

birkirb:
Yup, same goes for the status code, which would usually be set by the action so it's imperative the compression check is executed after that.

brendensoares:
Ordering is up to the developer and trying to force it to be last breaks the convention that ordering is configured outside of each filter.

I've created a new issue to revisit this topic in the near future.

Thanks for the contribution @birkirb!

PR #894, opened 2015-03-12
UI and functional enhancements to the test runner page

eric-gilbertson:
enhanced-testrunner
   - more condensed layout to reduce need to scroll
   - move buttons to left so they are near the test name
   - add file level run, e.g. all tests in a given suite
   - remove table striping to better show test groups

## TODO
   - [ ] display tests in dir hierarchy, e.g. controller, service, model
   - [ ] add show/hide for each test dir
   - [ ] display sorted by name

### DIFF
--- a/modules/testrunner/app/views/TestRunner/Index.html
+++ b/modules/testrunner/app/views/TestRunner/Index.html
@@ -10,9 +10,12 @@
 		<script src="{{url `Root`}}/@tests/public/js/highlight.pack.js" type="text/javascript"></script>
 		<style>
 		header { padding:20px 0; background-color:#ADD8E6 }
+		button.file-test { margin-bottom: 4px; margin-left: 6px }
+		table.tests tr { border-bottom: 1px solid #ddd; background-color: #f9f9f9; }
+		.table > tbody > tr > td { vertical-align: middle; }
 		.passed td { background-color: #90EE90 !important; }
 		.failed td { background-color: #FFB6C1 !important; }
-		.tests td.name, .tests td.result { padding-top: 13px; }
+		.tests td.name, .tests td.result { }
 		pre { font-size:10px; white-space: pre; }
 		.name { width: 25%; }
 		.w100 { width: 100%; }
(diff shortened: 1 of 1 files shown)


anonx:
Hi, @eric-gilbertson. I like your ideas: https://groups.google.com/forum/#!topic/revel-framework/mt__Z4w4a-E
Actually one of my early contributions to Revel was #725 and I was thinking about these things too. So, :+1: to your PR.

I prefer buttons at the right side though. I usually do not use mouse but when I do my cursor is at the right side of the screen most of the time. What are the preference of other users? I know `testrunner` is not likely the most popular part of Revel. But probably somebody will share with us their opinion regarding this aspect? @pedromorgan, @brendensoares, @notzippy 

P.S.: testrunner has been moved to a `modules` repo. Thus, you need to target it there: https://github.com/revel/modules/tree/develop

pedromorgan:
- Buttons on the right side I would really like..
- Also a row hover, so I can see which line i'm on..

I also wouldnt mind the whole thing being more compact, with less spacing and smaller fonts, so i can get more on a page

just my thoughts

eric-gilbertson:
As to button placement, one solution would be to place them on both the left and the right. I think it is good to have them on the left since it provides a tight visual coupling between then test name and the run button. Also, it accommodates those who work out of page left. 

Row hover tooltips could be added but given the use of bootstrap styling the rows cannot be shrunk much more as they already use the "xs" buttons.

If there is interest on this approach I will check it into the modules repo.

anonx:
@eric-gilbertson What if test name would be a link for running tests. And we can make it `display: block` and `width: 100%`.

eric-gilbertson:
I did try replacing the buttons with simple links and it does result in the saving of few more pixels per row. OTOH, I thought the presentation was less clear and that it was best to use a button because the click was invoking an action (as opposed to a view). Note that the condensed layout I checked in does already result in a vertical saving of approximately 30%. My $0.02 would be to keep the buttons for this iteration, but I can change back to links if that is the consensus. I would like to finish this issue so please let me know if you would like this change pushed to into modules and whether to use buttons or links. 

-Eric

notzippy:
Another two cents.. If you hid the subtests for each controller you would save a lot more space, add a + beside the test button for the controller to display / hide the subtests.

brendensoares:
@notzippy this is a good idea. Some more:
1. Show a live, [stacked](http://getbootstrap.com/components/#progress-stacked) [progress bar](http://foundation.zurb.com/docs/components/progress_bars.html)(linked to foundation framework because I like its markup better than bootstrap :)) of test results
2. Add animations to indicate status (e.g. blinking green when running)

brendensoares:
@anonx why did we close this?

PR #899, opened 2015-03-18
Fix broken fake app tests

anonx:
#855 revealed some problems with our fake app tests. Particularly, incorrect order of elements' initialization. This PR fixes broken tests.

There were two options: either to implement every of booking sample controllers + job module controller in out fake app. Or just move some booking sample's files to testdata directory and use them there. I have settle on the second option.

Now tests of revel/revel do not depend on revel/samples, only on revel/modules.

@brendensoares What else do we need for v0.12 release?

### DIFF
--- a/fakeapp_test.go
+++ b/fakeapp_test.go
@@ -5,7 +5,6 @@ import (
 	"log"
 	"os"
 	"path"
-	"path/filepath"
 	"reflect"
 )
 
@@ -59,7 +58,7 @@ func (c Static) Serve(prefix, filepath string) Result {
 }
 
 func startFakeBookingApp() {
-	Init("prod", "github.com/revel/samples/booking", "")
+	Init("prod", "github.com/revel/revel/testdata", "")
 
 	// Disable logging.
 	TRACE = log.New(ioutil.Discard, "", 0)
@@ -67,12 +66,6 @@ func startFakeBookingApp() {
 	WARN = TRACE
 	ERROR = TRACE
 
-	runStartupHooks()
-
-	MainRouter = NewRouter("")
-	routesFile, _ := ioutil.ReadFile(filepath.Join(BasePath, "conf", "routes"))
-	MainRouter.Routes, _ = parseRoutes("", "", string(routesFile), false)
-	MainRouter.updateTree()
 	MainTemplateLoader = NewTemplateLoader([]string{ViewsPath, path.Join(RevelPath, "templates")})
 	MainTemplateLoader.Refresh()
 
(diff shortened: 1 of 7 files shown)


anonx:
@brendensoares #728 changed the behaviour of testrunner.

For demonstration, here is a sample action:
```go
// ToDo redirects user to Index page.
func (c *App) ToDo() revel.Result {
	// Redirecting to App.Index.
	return c.Redirect(routes.App.Index())
}
```

Before #728 we had:
```go
func (t *AppTest) TestThatToDoWorks() {
	t.Get(routes.ToDo())
	// Status is equal to the status of App.Index (destination page).
	// Body is equal to the body of App.Index (destination page).
}
```

After merging we have:
```go
func (t *AppTest) TestThatToDoWorks() {
	t.Get(routes.ToDo())
	// Status is equal to 301 Moved Permanently.
	// Body is empty.
}
```

anonx:
#900 replaces "301 Moved Permanently" by our good old friend "302 Moved Temporarily". Combination of #899 and #900 will fix the broken tests.

brendensoares:
Looking to merge this now and release v0.12. Our docs GH Pages repo has already been merged.

anonx:
@brendensoares, @pedromorgan Can we make some changes to our v0.12 docs:
* Update version on index page.
Revel Version
* Remove a dot or move it to the end of "Read how the chat is implemented" sentence.
* Remove `get revel framework` and `go get github.com/revel/revel`. There is no need to do it. Revel will be automatically go getted as a dependency of `revel/cmd/revel`.
Index Page

anonx:
Yeah, and can we have more elegant **h1**'s? I'm mean a bit smaller and with a smaller bottom margin (margin between `Quick Start` text above and a horizontal line).

+ `Quick Start` link at the main menu leads to mostly empty pages (with some code fragments without code highlighting). Probably we should remove it for now? And when we have something to say we will return it back?

Apart from that I like our new docs.

anonx:
Oh, my mistake. Those pages are not empty. Only `Quick Start / Introduction / ...` ones. So, can we remove just empty links and make menu consistent with the rest of revel web-site (I mean left side menu).

* One of empty `Quick Start / Intro` pages. Pay attention to the menu:
Quick start page

* Left side menu which is used for other sections of web-site such as Manual, Tutorial, etc. Looks more professional, doesn't it?
Left side menu

brendensoares:
Good catches @anonx. @pedromorgan is taking ownership of the docs. I agree that the styling can be improved.

@pedromorgan can you fix the above 3 items noted? After that, we can resync `develop` branch with `master` via `git pull` and move forward with additional changes in `develop`. If we need to make patch releases to master in between major releases, then that's fine, but let's still use our version control workflow.

Issue #430, opened 2013-12-30
Controller Redirect doesn't use the correct HTTP status codes


rickb777:
In results.go, there are two places where http.StatusFound (302) is used. This happens to work because browsers support it for backward compatibility, but is not to the current standard HTTP 1.1 behaviour (http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3).

HTTP provides 303 and 307 with slightly different semantics. The 302 code is effectively deprecated, usually being treated as if it were 303.
- With a Post/Redirect/Get request, a See Other 303 is expected.
- With a simple redirect, a Temporary Redirect 307 is expected.

Correct use of these improves the browser behaviour in subtle but useful ways, improving the user experience.

Therefore, the revel.Controller needs to be two methods (or one with another parameter). The existing Redirect method should send a 307 rather than 302. A new method ("SeeOther" perhaps) should send a 303 but otherwise be quite similar.

notzippy:
Fixed

PR #900, opened 2015-03-18
Use "302 Found" HTTP code for redirect

anonx:
#728 introduced "301 Permanent Redirect" for `Redirect` result. But the way we use `Redirect` method requires "302 Found" aka Moved Temporarily. This PR restores the status code we used before.


alkchr:
It should be replaced by `http.StatusSeeOther` (status code 303) [[1]]. #430 is related.

[1]: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection

PR #901, opened 2015-03-22
Assets pipeline

huacnlee:


brendensoares:
We need to make sure that asset pre-processing is pluggable. This PR uses node.js and external tools to compile assets, but I would much rather not introduce such dependencies.

Here are so native Go libiraries:

- HTML
  - [HAML](http://haml.info/docs/yardoc/file.REFERENCE.html)
    - https://github.com/dddaisuke/go-haml
    - https://github.com/eknkc/amber templating inspired by haml/jade
  - [Slim](http://www.rubydoc.info/gems/slim/frames)
    - https://github.com/yosssi/ace templating inspired by slim/jade
- CSS
  - [SASS](http://sass-lang.com/documentation/file.SASS_REFERENCE.html)
    - https://github.com/moovweb/gosass
- Javascript
  - [CoffeeScript](http://coffeescript.org/)
    - https://github.com/jashkenas/coffeescript/tree/master/src (we would need to port this logic to Go)

We can start by using external commands, but I would really like to use native code and we can transition to this goal gradually if the asset pipeline is modular.

I can start an RFC to define these points and use this PR as a starting point/reference.

huacnlee:
Sass looks like to be simply change to use Go lib, because there have libsass, But CoffeeScript is hard to do that.

And `train` can compile asset files to `public/assets`, so we don't need install Node.js in production env.

huacnlee:
What means RFC?

brendensoares:
@huacnlee  www.github.com/revel/rfc

I see what you mean that production doesn't require it. Sounds good. I'm happy to merge this into `develop` and build on it.

@revel/core thoughts?

huacnlee:
https://github.com/huacnlee/mediom

There have a example app use this feature.

huobazi:
+1

brendensoares:
@huacnlee thanks for the sample app!

runner-mei:
+1

brendensoares:
Looking at this again, why add this directly to Revel when you could just use the node tools? There are many asset precompiler tools in use these days.

I'm leaning toward not adding this to Revel. @revel/core thoughts?

notzippy:
Closing, at the moment we are focusing on rebuilding core components - thanks for the PR

PR #903, opened 2015-03-22
v0.12 Release

brendensoares:


anonx:
@brendensoares Don't forget to include in release notes info on how to upgrade to v0.12. Looks like testing package is the only breaking change. So, users will have to update import paths of `TestSuite` type in files of `tests` directory.
* Old version:
```go
package tests

import "github.com/revel/revel"

type AppTest struct {
	revel.TestSuite
}
```
* Required change:
```go
package tests

import "github.com/revel/revel/testing"

type AppTest struct {
	testing.TestSuite
}
```
Apart from that everything works for my not-so-small application as expected, nothing is broken, its tests are passed.

brendensoares:
@anonx I noted the new import path. Also updated the `README.md` and switched the TravisCI branch to `master`.

We need to figure out a clean way to switch the Travis branch during release. For now, we'll just have to switch it back to `develop` after the release. Any better ideas?

anonx:
@brendensoares We didn't update `version.go`: https://github.com/revel/revel/blob/master/version.go#L5

I'll let you know when I come up with a better idea.

brendensoares:
@anonx Should it be `0.12.0` or `0.12.0rc` or what?

notzippy:
@brendensoares  I would go with all numeric `0.12.0`

anonx:
@brendensoares I thought `v0.12.0dev` meant it was `develop` branch. Well, if it's not and `dev` part is intentional meaning it's not stable then it's OK (although, I think `v0` means exactly that without any suffixes).

brendensoares:
@anonx I agree that v0 means we've yet to reach the desired foundational features.

@notzippy I'm good with plain `0.12.0`, but we may want to discuss a release candidate workflow to make releases more often as @pedromorgan has recently asked.

brendensoares:
@anonx `BUILD_DATE` refers to date of release in my mind. Thoughts? I've updated both date and version number now.

PR #915, opened 2015-03-29
Controller add c.Finish method to allow user give a Result and break off action call stack.

huacnlee:
Useful method to give a Result and break off action call stack.

### For example:

before:

```go
func (c Posts) requireUser() revel.Result {
  if c.Session["user_id"] == nil {
    return c.Redirect("/login")
  }
  return nil
}

func (c Posts) Edit() revel.Result {
  if r := c.requireUser(); r != nil {
    return r
  }

  posts := DB.FindPosts()
  c.RenderArgs["posts"] = posts
  return c.Render()
}
```

after:

```go
func (c Posts) requireUser {
  if c.Session["user_id"] == nil {
    c.Finish(c.Redirect("/login"))
  }
}

func (c Posts) Edit() revel.Result {
  c.requireUser()
  // Follow code will not run, if requireUser called c.Finish
  posts := DB.FindPosts()
  c.RenderArgs["posts"] = posts
  return c.Render()
}
```
 
- No login: [GET /posts/1/edit] -> [Require user] -> [Redirect] -> [End]
- Logined:  [GET /posts/1/edit] -> [Require user] -> [FindPosts] -> [Render] -> [End]

----

PS: Where is the controller test cases?

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -70,6 +70,32 @@ func (c *Controller) setStatusIfNil(status int) {
 	}
 }
 
+// Useful method to give a Result and break off action call stack.
+//
+// For example:
+//
+//   func (c Posts) requireUser {
+//     if c.Session["user_id"] == 0 {
+//       c.Finish(c.Redirect("/login"))
+//     }
+//   }
+//
+//   func (c Posts) Edit() revel.Result {
+//     c.requireUser()
+//     // Follow code will not run, if requireUser called c.Finish
+//     posts := DB.FindPosts()
+//     c.RenderArgs["posts"] = posts
+//     return c.Render()
+//   }
+//
+// - No login: [GET /posts/1/edit] -> [Require user] -> [Redirect] -> [End]
+// - Logined:  [GET /posts/1/edit] -> [Require user] -> [FindPosts] -> [Render] -> [End]
+//
+func (c *Controller) Finish(r Result) {
+	r.Apply(c.Request, c.Response)
+	panic(nil)
+}
+
 // Render a template corresponding to the calling Controller method.
 // Arguments will be added to c.RenderArgs prior to rendering the template.
 // They are keyed on their local identifier.


notzippy:
Normally a filter is used to implement this behavior, can you think of a
situation that this would preferable over a filter?

huacnlee:
RequireUser just an use case.
Use Filter to do this it too complex, and I do even want to let all action don't need return revel.Result like this:

```go
func (c Posts) Index() {
    c.requireUser()
    if !IsMember() {
       // When call Redirect, redirect and stop here.
       c.Redirect("/")
    }
    topics := DB.FindTopics()
    c.RenderArgs["topics"] = topics
    // Call render, and apply response
    c.Render()
}
```

## Function list

- Render
- Redirect
- RenderError
- RenderTemplate
- RenderJson
- RenderJsonP
- RenderXml
- RenderText
- RenderHtml
- NotFound
- Forbidden
- RenderFile
- RenderBinary

anonx:
@huacnlee By the way, this is a very interesting point.

@revel/core What are the pros and cons of using `return c.Smth()` rather than just `c.Smth()`? Absence of `return` would make `Render` method even more magic, wouldn't it? But on the other hand, it would allow to use such constructions as `requireUser` (from code fragment above).

Although, for `requireUser` case, interceptors are probably a better approach:
```go
func (c *App) RequireAuth() revel.Result {
	if _, ok := c.Session["email"]; ok {
		return nil
	}
	return c.RenderTemplate("accounts/login.html")
}

func (c *App) Index() revel.Result {
	// ToDo
	return c.Render()
}

func init() {
	revel.InterceptMethod((*App).RequireAuth, revel.BEFORE)
}
```
If user is not authorized he/she will see a login page. And it will work for any action of `App` controller.

notzippy:
@anonx Wouldn't that need to be `c.Smth(revel.Result)` and in that case it is less obvious what the the action is supposed needs to do. Unless I am missing your point here.

huacnlee:
When I call `Redirect`, that means sure are I want to redirect page to another page, why do I need to do return?

I know that use return will be let code readability, But why do must be return a `revel.Result` type?

requireUser just a little use case

```go
// Before
func (c Posts) isOwner(t Post) revel.Result {
    if t.UserId != c.currentUser.id {
        return c.Forbidden("This is not yours.")
    }
    return nil
}

func (c Posts) Edit() revel.Result {
    p, err := DB.FindPost(c.Params.Get("id"))
    if r := c.isOwner(p); r != nil {
        return r
    }

    // Do render
    return c.Render("posts/edit.html")
}


// After 
func (c Posts) isOwner(t Post) {
    if t.UserId != c.currentUser.id {
        c.Forbidden("This is not yours.")
        return
    }
    nil
}

func (c Posts) Edit() {
    p, err := DB.FindPost(c.Params.Get("id"))
    c.isOwner(p)

    // Do render
    c.Render("posts/edit.html")
}
```

nadeemc:
Just chiming in here. I vote with @notzippy that the use case here is for a filter. I implement something similar when providing a user object to my controllers.

This is an example filter.go file:
```go
func init() {
	revel.InterceptFunc(VerifyAuth, revel.BEFORE, &Dashboard{})
}

func VerifyAuth(c *revel.Controller) revel.Result {
	authId := getCookieValue(c, ".auth")

	var u User
	var err error
	if authId == "" {
		err = errors.New("AuthId is not set.")
	} else {
		u, err = users.GetByAuthId(authId)
	}

	if err != nil {
		if isAjaxRequest(c) {
			return cr.RenderStatusCode(http.StatusUnauthorized)
		} else {
			return c.Redirect("/accounts/login/")
		}
	}

	user := views.LayoutUser{
		UserId:     u.UserId,
		AuthId:     authId,
		Name:      u.FirstName,
		Email:       u.EmailAddress,
	}
	c.RenderArgs["User"] = user
	c.Params.Set("authId", authId)

	return nil
}

func getCookieValue(c *revel.Controller, cookieName string) string {
	cookie, err := c.Request.Cookie(cookieName)
	if err == nil {
		return cookie.Value
	}
	return ""
}

func isAjaxRequest(c *revel.Controller) bool {
	acceptHeader := c.Request.Header.Get("Accept")
	return acceptHeader == "" ||
		strings.Index(acceptHeader, "*/*") == 0 ||
		strings.Contains(strings.ToLower(acceptHeader), "javascript") ||
		strings.Contains(strings.ToLower(acceptHeader), "json")
}
```

This would then be the corresponding dashboard.go controller:
```go
type Dashboard struct {
	*revel.Controller
}

func (c Dashboard) Index(authId string) revel.Result {
	// Access authId value from filter using "authId" parameter

       // Access full user object
      user := c.RenderArgs["User"].(views.LayoutUser)
}
```

While using RenderArgs isn't the most ideal approach, this does do what you're asking.

I think a separate feature though would be to allow InterceptFunc() to not be called on some controller methods, i.e. AllowAnonymous()

For the time being to get around that, I use two controllers and then manually configure routes as needed.

brendensoares:
@huacnlee I think this is a simple solution and I am tempted to merge it, but some good points have been made here about avoiding magic via conventions (e.g. this method will actually end the request if you call it). It's better/safer to explicitly return a value.

As far a a user auth filter, I would actually recommend a [`BEFORE` interceptor](https://revel.github.io/manual/interceptors.html) myself.

I'm going to close this PR for now for these reasons. Thanks for contributing!

Issue #505, opened 2014-02-14
Using objects for filesystem paths instead of strings


ghost:
Can we use something more elaborate for storing filesystem paths, to abstract operating system differences? Here's my try at porting the python's [pathlib](http://docs.python.org/3.4/library/pathlib.html) module (see also [PEP-0428](http://www.python.org/dev/peps/pep-0428/)):
https://github.com/opennota/pathlib

brendensoares:
@anonx @pushrax this is related to #260 right?

anonx:
@brendensoares Right, and it is labeled for consideration in v0.12.
I personally have never felt comfortable concatenating paths and making that kind of stuff. That's too fragile. So, I'm :+1: the idea of using some abstraction layer instead. Haven't investigated whether pathlib fits our needs or something else is required, though.

brendensoares:
#270 looks related

Issue #260, opened 2013-08-06
Slashes should be normalized in paths


martinml:
_**tl;dr**: Windows uses `\` as a path separator and that messes up `BasePath` and possibly others_

While reading docs and source I kept reading that Revel included a `PanicFilter` that would create nice error pages in case of `panic()`, but it wasn't working for me: in dev or in prod mode, I was seing always the raw stacktrace being dumped to my browser.

Perusing the source I arrived to `findRelevantStackFrame()` in `errors.go`, which I guess tries to find out if the panic was inside the application by finding `BasePath` in the stack trace. The problem is that if you use Windows, it never finds it, because the slashes are wrong:

 - stack = `(...) C:/gocode/src/myapp/app/controllers/whatever.go (...)`
 - BasePath = `C:\gocode\src/myapp`

Note the mixing of different slashes. Windows accepts both so everything works, but if you search `BasePath` in `stack`, it won't find it.

A solution might be replacing in `SourcePath` all the backslashes (`\`) with forward slashes (`/`) _but only if we are in Windows_, but that seems a bit hacky. What would be a better solution?

jeevatkm:
#1028 PR is merged to `develop` branch for handling path separator using `filepath.ToSlash` in `error.go`. So I'm closing this one.

PR #928, opened 2015-04-23
Fix function not working on Windows because of backslash

JaTochNietDan:
So it turns out us Windows folk have been missing out on very nice detailed information in the testrunner module. All we would get is a red bar with no information when a test failed....we thought this was normal until a Linux user pointed out the fact that he had all of this information on why a test failed inline.

This fixes the issue, problem is the base path on Windows looks like `C:\Gopath\src/github.com/blabla/blabla`. Converting all of the backslashes to forward slashes fixes this fully :).

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -107,6 +107,7 @@ func (e *Error) ContextSource() []sourceLine {
 // Return the character index of the first relevant stack frame, or -1 if none were found.
 // Additionally it returns the base path of the tree in which the identified code resides.
 func findRelevantStackFrame(stack string) (int, string) {
+	BasePath = strings.Replace(BasePath, "\\", "/", -1)
 	if frame := strings.Index(stack, BasePath); frame != -1 {
 		return frame, BasePath
 	}


ychartois:
Much better like that ;)

brendensoares:
I marked this accepted, but I wonder if this is the right solution.

This is the same issue as #260.

@revel/core comments?

notzippy:
Personally I would do the following, it should give a OS independent feel

    BasePath = filepath.Join(strings.Split(BasePath,`\`))

anonx:
@brendensoares This is probably not the right way. But if we are not going to introduce the right one in v0.13 let's use this workaround.
Does anybody want to start working on implementation of right solution?

pedromorgan:
Qt Seem to have it sussed, and use "/" internally.. for all filePath

and two fucnctions toNativeSeperators(), fromNativeSeperators()

http://doc.qt.io/qt-4.8/qdir.html#fromNativeSeparators
http://doc.qt.io/qt-4.8/qdir.html#toNativeSeparators

anonx:
BTW there is a ``filepath.ToSlash`` equivalent (to `strings.Replace(s, "\\", "/", -1)`) in standard `path/filepath` library that transforms your path no matter what separators it uses to Unix looking path format with slashes.

brendensoares:
Deferring this as it is low impact/probability and we know of a better solution. Just need to do it.

## TODO

- [ ] implement platform independent path separators

alkchr:
@brendensoares What is a better solution in your opinion? I've used different approaches for my framework including #505 like. But ended up using standard `filepath.Join`, `filepath.ToSlash`, `filepath.FromSlash`, and 3 simple [helper functions](https://github.com/colegion/goal/blob/master/utils/path/path.go).

jeevatkm:
I believe we can close this PR in-favor of #1028

alkchr:
:+1: Yeah, absolutely. But somebody will have to retarget #1028 to `develop`.

jeevatkm:
#1028 merged to `develop` branch, I'm closing this one.

PR #935, opened 2015-05-19
Add Order to OnAppStart

vahid-sohrabloo:
Add Order To onAppStart.
With This change we can order to run function in runStartupHooks.

### DIFF
--- a/server.go
+++ b/server.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


vahid-sohrabloo:
@anonx  I added test. Why you add needs testting tag?

anonx:
@vahid-sohrabloo, thank you for this PR. Everything is OK, "needs testing" is a label for @revel/core to make sure everything is correct and to double check things.

vahid-sohrabloo:
@anonx Thank you

server.go
```diff
@@ -131,12 +132,32 @@ func Run(port int) {
...
 	for _, hook := range startupHooks {
-		hook()
+		hook.f()
 	}
 }
 
-var startupHooks []func()
+type StartupHook struct {
+	order int
+	f     func()
```
brendensoares: Not sure I like such a vague variable name on a `struct`. Thoughts?

brendensoares:
Overall, good enhancement.

server.go
```diff
@@ -174,6 +195,10 @@ var startupHooks []func()
 // This can be useful when you need to establish connections to databases or third-party services,
 // setup app components, compile assets, or any thing you need to do between starting Revel and accepting connections.
 //
-func OnAppStart(f func()) {
-	startupHooks = append(startupHooks, f)
+func OnAppStart(f func(), order ...int) {
+	o := 1
```
notzippy: I think the default value should be 5 and we have some sort of guideline as to the range of values to be used.
ie  Default / normal is 5, high / first is 1, low / last is 10. 
So people know that revel builtin startuphooks will run at 5, if they want something before or after it should be higher or lower.

brendensoares: Agreed.

vahid-sohrabloo: I'm not Agree. people can use negative numbers.

server.go
vahid-sohrabloo: maybe. do you have name suggest?

brendensoares: `hook.callback()` seems appropriate.

server.go
brendensoares: That's true. We can assume all hooks will run at the same level just as they do currently and can be changed to different priority if needed. Should be good for backward compatibility.

PR #939, opened 2015-05-26
Fix for issue #936 suggested by @anonx

bejayoharen:
Fix for issue #936 suggested by @anonx

Issue #936, opened 2015-05-20
Large uploads fail


bejayoharen:
When doing a large upload, I get a timeout if the upload takes more than 1 minute -- even if the client is steadily providing data. The server should not timeout if while the client is providing data.

The error I see in my logs comes from here (line 49 in params.go):

		if err := req.ParseMultipartForm(32 << 20 /* 32 MB */); err != nil {
			WARN.Println("Error parsing request body:", err)

In revel/server.go, I noticed the server timeouts are set, so I was able to work around my issue with:

    init() {
    	revel.OnAppStart(fixServerTimeout)
    }

    func fixServerTimeout() {
    	revel.Server.ReadTimeout = 0
    	revel.Server.WriteTimeout = 0
    }

which is neither documented or intuitive, but does work.

I think your intention may have been to timeout if the client does not provide data for a minute, or if the server fails to respond after a minute.

Consider, however, the case where the client is uploading a huge file: the server does not respond for a minute because it is still reading, which is perfectly legitimate. However, since it's not writing for a minute, it times out because of the write timeout (Another workaround might be to use threading and send continues).

I don't think the server read/write timeouts should be set at all, or if they are, they should be larger and configurable in the conf file.

anonx:
> I don't think the server read/write timeouts should be set at all, or if they are, they should be larger and configurable in the conf file.

Agree. Looks like the timeouts became a reason of more bad things than good. Let's use "no timeouts" option by default and make it configurable. The following change is required (https://github.com/revel/revel/blob/develop/server.go#L85-L90):

```diff
Server = &http.Server{
	Addr:         localAddress,
	Handler:      http.HandlerFunc(handle),
-	ReadTimeout:  time.Minute,
-	WriteTimeout: time.Minute,
+	ReadTimeout:  Config.IntDefault("timeout.read", 0) * time.Second,
+	WriteTimeout: Config.IntDefault("timeout.write", 0) * time.Second,
}
```
@bejayoharen Do you want to create a PR?

bejayoharen:
Okay, I just copied your changes -- no testing. I don't really know how the revel config system works internally.

anonx:
@bejayoharen Yeah, I didn't test either and missed the fact `int` has to be type casted to `time.Duration`.

bejayoharen:
Okay: https://github.com/revel/revel/pull/940

anonx:
:+1: Thanks.

PR #940, opened 2015-05-26
Fix for issue #936 suggested by @anonx (#939)

bejayoharen:


anonx:
Thank you, @bejayoharen. It has been merged into develop (see last two commits [here](https://github.com/revel/revel/commits/develop)).

@revel/core New behaviour is introduced by this PR. There are no timeouts by-default. However, if somebody needs them he/she can use `timeout.read` and `timeout.write` options in their `config/app.conf`. Example:
```ini
# Timeout specifies a time limit for request (in seconds) made by a single client.
# A Timeout of zero means no timeout.
timeout.read = 300
timeout.write = 60
```

brendensoares:
@anonx would it be better to have the default be a sane value? No timeout seems dangerous for a default.

bejayoharen:
I think 0 is the best default:

1. The timeouts are **very** unintuitive. For example, if you set the write timeout to 1 minute, then the server will timeout if it takes longer than one minute to *upload* a file. This is because, in the meantime, while uploading, nothing has been written to the socket, so it closes the connection.
2. Even large values could interfere with connections used for things like long-polling. (Why close a connect used for long-polling just because no data has gone through if you are still getting TCP keep-alives?) I think 0 is very sensible and sane for this use-case as well.

I think it's better to not use defaults and instead rely on lower-level TCP timeouts, file size limits and so on.

brendensoares:
@bejayoharen time will tell what is the bet solution, but I feel if we make it clear that file uploads need a 0 timeout, then it wouldn't be bad to have safe defaults, like 60 seconds. Having no timeout makes a web app more vulnerable to DoS attacks, on purpose or on accident (high traffic spikes).

anonx:
@brendensoares, Even if we decide to use some kind of default timeouts this PR is useful. It allows us to use `conf/app.conf.template` for that rather than hardcode something to `server.go`. So timeout values will be documented.

bejayoharen:
@brendensoares For DoS, I would suggest adding the ability to set a minimum data rate, configure TCP timeouts, as well as the ability to limit request sizes (perhaps separate limits for both header and body), instead of socket read/write timeouts.

PR #943, opened 2015-05-30
always use HttpOnly and default to Secure cookies

### LINKED ISSUE #942: default HttpOnly and Secure to true
Leaving cookie.httponly and cookie.secure to false by default on cookies will cause folks without detailed knowledge of how web security works (which is most of folks) wide open to impersonation, replay, and XSS attacks.

To handle developers trying to work locally, a check can be done to see if we are in DevMode and to allow the default to be false then.

jmhodges:
Always use Secure cookies (that is, require HTTPS for the cookie to be
used) unless we're in dev mode.

Always set the session cookies as HttpOnly so that XSS attacks can't dig
into them. They'll still be set on XHR requests, but the JavaScript
code won't see it.

Closes #942

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
   - redis-server
 install:
   - export PATH=$PATH:$HOME/gopath/bin
-  - export REVEL_BRANCH="master"
+  - export REVEL_BRANCH="develop"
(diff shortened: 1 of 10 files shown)


jmhodges:
Ping!

jmhodges:
I've fixed up the branching issue.

DavidHunter:
I think this is suppose to target the develop branch

PR #985, opened 2015-09-23
Add InitDirectly to initialize with BasePath

ianterrell:
Some deployment environments (Heroku) do not put the Revel application
underneath a normal src directory as you might expect in a full
Go environment setup.

This allows Revel to initialize itself directly with a BasePath
as well as the path to the src path where Revel exists (in a deployment,
perhaps ./Godeps/workspace or ./vendor)/

It is arguable that the currently named `InitDirectly` should be the primary `Init`, and the existing `Init` should be renamed to `InitWithImportPath` or something else.

Combined with generating a new main.go, this change allows easy deployment to Heroku from source with the default go buildpack maintained by Heroku.

cc @brendensoares


ianterrell:
Requested pull into Master by default and accident. Closing in favor of https://github.com/revel/revel/pull/986.

PR #994, opened 2015-10-05
 Fix change Server.Addr in the hook function and it does not work.

runner-mei:
Fix change Server.Addr in the hook function and it does not work.
Why change Server.Addr?
I hope read port from another config file, the config file will modified by user and don't overrided while reinstall or update my software.

### DIFF
--- a/server.go
+++ b/server.go
@@ -110,7 +110,7 @@ func Run(port int) {
 
 	go func() {
 		time.Sleep(100 * time.Millisecond)
-		fmt.Printf("Listening on %s...\n", localAddress)
+		fmt.Printf("Listening on %s...\n", Server.Addr)
 	}()
 
 	if HttpSsl {
@@ -122,7 +122,7 @@ func Run(port int) {
 		ERROR.Fatalln("Failed to listen:",
 			Server.ListenAndServeTLS(HttpSslCert, HttpSslKey))
 	} else {
-		listener, err := net.Listen(network, localAddress)
+		listener, err := net.Listen(network, Server.Addr)
 		if err != nil {
 			ERROR.Fatalln("Failed to listen:", err)
 		}


alkchr:
+1 for this change. It doesn't break things, easy, but may be useful if you want to modify address from a [hook](https://github.com/runner-mei/revel/blob/patch-1/server.go#L92) function.

Though, per [contributing guideline](https://github.com/revel/revel/blob/master/CONTRIBUTING.md#how-revel-is-developed) `develop` branch must be targeted not `master`.

brendensoares:
@runner-mei `localAddress` is derived from `HttpAddr` which is configurable based on `http.addr` in the config file. Why would you add a new config param?

@alkchr @revel/core  am I crazy?

alkchr:
@brendensoares Yeah, you're right. It is configurable. But IIRC somebody had a use-case when they wanted to modify the addr dynamically from startup hooks.
```go
func Run(port int) {
	address := HttpAddr // Here we are initializing it.

	// Creating a server here.
	Server = &http.Server{
		Addr:         localAddress, // This is the address from the line above.
		...
	}

	// The startup hooks change the Server.Addr
	// based on some dynamic conditions.
	runStartupHooks()

	// But the change doesn't have affect if we're not using SSL.
	if !HttpSsl {
		net.Listen(network, localAddress)
	} else {
		Server.ListenAndServeTLS(...) // But it affects this line.
	}
}
```
So, @runner-mei is not adding a new parameter. He makes the code consistent between HTTPS and HTTP modes.

jeevatkm:
My understanding is existing flow stay as-is with configurable values like @brendensoares mentioned and using `Server.Addr` in `net.Listen(network, Server.Addr)` provide ability to modify via startup hooks since it is exported variable `revel.Server.Addr` and startup hooks cannot have access to un-exported/local variable like `localAddress` from revel package.

BTW #999 points to correct branch for this change.

PR #995, opened 2015-10-08
Use improved deflate/gzip compressor.

klauspost:
This pure-Go library (by yours truly) has optimized gzip/deflate function. At default compression levels this gives ~70% improvement in throughput for web content, as well as fewer allocations.

For performance comparisons see the package and https://blog.klauspost.com/gzip-performance-for-go-webservers/

I would love to give revel-specific benchmarks, but the benchmarks in "compress_test.go" appear broken, since it doesn't seem like compression is applied at all.

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -1,12 +1,13 @@
 package revel
 
 import (
-	"compress/gzip"
-	"compress/zlib"
 	"io"
 	"net/http"
 	"strconv"
 	"strings"
+
+	"github.com/klauspost/compress/gzip"
+	"github.com/klauspost/compress/zlib"
 )
 
 var compressionTypes = [...]string{


alkchr:
@klauspost :+1: Well done and the blog post is great.

brendensoares:
This sounds exciting. I'd like to run some tests before we merge. I'd also like to see if there is a problem that I can fix in the compress_test.go that you mention. I'm unaware of any specific bugs there.

brendensoares:
@klauspost the `compress_test.go` benchmarks work for me:
```
$ go test -bench=.
PASS
BenchmarkRenderCompressed	   30000	     42463 ns/op
BenchmarkRenderUnCompressed	   30000	     41268 ns/op
...
```

There was an error in `BenchmarkInvoker`, but that seems unrelated here.

klauspost:
@brendensoares - the "BenchmarkRenderCompressed" doesn't appear to apply compression, so you are not benchmarking compression ;)

brendensoares:
@klauspost How do you know it's not applying compression?

When I test this PR I get an improvement:
```
BenchmarkRenderCompressed	   30000	     40650 ns/op
BenchmarkRenderUnCompressed	   30000	     41156 ns/op
```

PR #999, opened 2015-10-13
 Fix change Server.Addr in the hook function and it does not work.

runner-mei:
Fix change Server.Addr in the hook function and it does not work.
Why change Server.Addr?
I hope read port from another config file, the config file is will modified by user and don't override while reinstall or update my software.

### DIFF
--- a/server.go
+++ b/server.go
@@ -110,7 +110,7 @@ func Run(port int) {
 
 	go func() {
 		time.Sleep(100 * time.Millisecond)
-		fmt.Printf("Listening on %s...\n", localAddress)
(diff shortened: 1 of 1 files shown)


runner-mei:
this is a pull request.

brendensoares:
Looks like a harmless change, but what is the benefit?

runner-mei:
My user change port of my app, but reinstall my app and will override config file, so I add hook to read port from the db.

jeevatkm:
@runner-mei Thanks for your response.

PR #1003, opened 2015-10-26
add Pluggable template engine support

runner-mei:
I want to use [pongo2](https://github.com/flosch/pongo2) in the revel project.
I redesign https://github.com/revel/revel/pull/506 and push it.

my samples project is https://github.com/runner-mei/revel_samples/tree/master/booking
my pongo2 module project is https://github.com/runner-mei/revel_modules/tree/master/pongo2/app

revel/cmd patch is  https://github.com/revel/cmd/pull/37

### DIFF
--- a/fakeapp_test.go
+++ b/fakeapp_test.go
@@ -67,7 +67,9 @@ func startFakeBookingApp() {
 	ERROR = TRACE
 
 	MainTemplateLoader = NewTemplateLoader([]string{ViewsPath, path.Join(RevelPath, "templates")})
-	MainTemplateLoader.Refresh()
+	if err := MainTemplateLoader.Refresh(); nil != err {
+		panic(err)
+	}
 
 	RegisterController((*Hotels)(nil),
 		[]*MethodType{
--- a/results.go
+++ b/results.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"runtime/debug"
 	"strconv"
 	"strings"
 	"time"
@@ -103,7 +104,6 @@ func (r ErrorResult) Apply(req *Request, resp *Response) {
 	} else {
 		websocket.Message.Send(req.Websocket, fmt.Sprint(revelError))
 	}
-
 }
 
 type PlaintextErrorResult struct {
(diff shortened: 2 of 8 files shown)


alkchr:
ping @brendensoares

xpbliss:
Good!
But the magic function, such as field , radio ,checkbox are modified  in module runner-Mei pongo2?

runner-mei:
Yes, such as field , radio ,checkbox are modified in module runner-Mei [pongo2](https://github.com/runner-mei/revel_modules/tree/master/pongo2/app).
   > my field function code is https://github.com/runner-mei/revel_modules/blob/master/pongo2/app/filters_field.go
  > my field function domo code is https://github.com/runner-mei/revel_samples/blob/master/booking/app/views/application/register.html#L14

Yes, the pongo2 also has a minor modifications , pull request is https://github.com/flosch/pongo2/pull/102

xpbliss:
Maybe it is not a good idea to modify the revel  and pongo2, it is not a pluggable template.
Whatever you do a good work!

alkchr:
@xpbliss I proposed a pluggable-everything architecture about half a year ago but it required quite a lot of work to rewrite the whole code base of Revel.
So, I started my own [project](https://github.com/colegion/goal) that is highly inspired by Revel but uses that new architecture. Key aspects are: everything is implemented as a set of independent tools rather than as a monolith, code generation instead of reflection, type safety, compatibility with the standard library, full customizability. It is still a work in progress and many things are missing. But what I like about it is the simplicity of adding new template engines, routers, etc. Probably, one day Revel team will adopt a similar approach.

notzippy:
I have always supported a pluggable template engine, but it should be configurable on a module basis, from looking at the code it appears you are switching globally. Is it possible to modify your code to accommodate this ?

runner-mei:
@notzippy Yes, template engine is switching globally. 
To my thinking, this is not a good idea, but it is possible That different templates are used in different modules. I add a [mixed template engine](https://github.com/runner-mei/revel/blob/revel_master/template_adapter_mixed.go) to support it.

brendensoares:
@runner-mei we're working to review, document, and release this in the next week.

Are you available to help with this process or should we take over this PR?

runner-mei:
What do I need to do?

brendensoares:
#1136 is taking over this PR. Let's collaborate there. Thanks @runner-mei !

PR #1007, opened 2015-11-19
Result implementation that allows custom content-types

peterkellyonline:
Ideal for file extensions

Use this to return a custom content-type
- revel currently only supports JSON, JSONP, XML, plain text and HTML
- If you set a content-type it will be overwritten using one of these Result types in revel.
- Safari, for example, will add a .txt to the end of your file name if you use plain text no matter what the file extension type is. e.g. returning a .eml file for download will be downloaded as .eml.txt by Safari - this fixes that issue

[Review on Reviewable](https://reviewable.io/reviews/revel/revel/1007)

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -159,8 +159,8 @@ func (c *Controller) RenderText(text string, objs ...interface{}) Result {
 	return &RenderTextResult{finalText}
 }
 
-// RenderText sets a custom content-type header, allowing flexibility of type returned e.g. various file extensions
-func (c *Controller) RenderText(contentType, text string) Result {
+// RenderCustom sets a custom content-type header, allowing flexibility of type returned e.g. various file extensions
+func (c *Controller) RenderCustom(contentType, text string) Result {
 	c.setStatusIfNil(http.StatusOK)
 
 	return &RenderCustomResult{


pedromorgan:
@peterkellyonline - What's the sue case of this (and a code snippet example) ?

peterkellyonline:
@pedromorgan Use case is when you want to return any Content-Type you want e.g. any from this list
http://www.freeformatter.com/mime-types-list.html#mime-types-list

A workaround in my code is to do the following

```
        c.Response.Out.Header().Set("Content-Disposition", "attachment;filename="test.eml")
	resp.WriteHeader(http.StatusOK, "message/rfc822; charset=utf-8")
	resp.Out.Write([]byte(c.Data))
```

But a generic Custom Render method taking a ContentType param would be great. I had honestly forgot about this PR :)

results.go
```diff
@@ -304,6 +304,17 @@ func (r RenderTextResult) Apply(req *Request, resp *Response) {
...
 }
 
+// CustomResultType - Use this to return a custom content-type e.g. file extension
+type RenderCustomResult struct {
+	ContentType string
+	Data        string
+}
+
+func (c RenderCustomResult) Apply(req *Request, resp *Response) {
+	resp.WriteHeader(http.StatusOK, c.ContentType+"; charset=utf-8")
```
jeevatkm: If we support this `RenderCustom` then `;  charset=utf-8` value have to come with `c.ContentType`, no hard-cording.

pedromorgan:
@peterkellyonline Can you target this at the development branch so we can get closer to merging. TIA.

notzippy:
Closing due to no response. Good idea so opened ticket #1130 for someone to implement

results.go
pedromorgan: Maybe an  optional third argument of encoding.. ?

PR #1010, opened 2015-11-25
Fixed getting number of arguments

oherych:
I can't explain why but this fix my problem. Without this changes I can't see any data if i sent it by Render(). I see this problem on my code and also on /@jobs too.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -87,7 +87,7 @@ func (c *Controller) Render(extraRenderArgs ...interface{}) Result {
 	c.setStatusIfNil(http.StatusOK)
 
 	// Get the calling function name.
-	_, _, line, ok := runtime.Caller(1)
+	_, _, line, ok := runtime.Caller(2)
 	if !ok {
 		ERROR.Println("Failed to get Caller information")
 	}


alkchr:
@DieselDraft What version of Go are you using? Do you have this issue with a skeleton app (that one that is generated by `revel new`)?

@revel/core If you're interested, [here](https://github.com/colegion/goal/blob/b04b2623423496ce098c93c86ff1378af88ea5cf/internal/skeleton/assets/handlers/app.go#L90) is what I'm doing to not depend on `runtime` and `reflect`. Simply saying: code generation. Every action name is known during build time, that's why we can generate an extra line of code for later use with `Render`.

DieselDraft:
Yes. It was created by `revel new`. I create new empty project i got same mistake

My Golang version
```
oherych@bukashka:~/go/src/bitbucket.org/bukorama$ go version
go version xgcc (Ubuntu 4.9.1-0ubuntu1) 4.9.1 linux/amd64
```


I added debug code
```
func (c *Controller) Render(extraRenderArgs ...interface{}) Result {
	c.setStatusIfNil(http.StatusOK)

	log.Println(runtime.Caller(0))
	log.Println(runtime.Caller(1))
	log.Println(runtime.Caller(2))
	log.Println(runtime.Caller(3))
...
```

And I get

```
2015/11/26 09:54:09 controller.go:90: 4413410 /home/oherych/go/src/github.com/revel/revel/controller.go 90 true
2015/11/26 09:54:09 controller.go:91: 4315294 /home/oherych/go/src/github.com/revel/modules/jobs/app/controllers/status.go 10 true
2015/11/26 09:54:09 controller.go:92: 4314767 /home/oherych/go/src/github.com/revel/modules/jobs/app/controllers/status.go 25 true
2015/11/26 09:54:09 controller.go:93: 140642171997375  0 true
```

/home/oherych/go/src/github.com/revel/modules/jobs/app/controllers/status.go  line:10
```
9 : type Jobs struct {
10:	*revel.Controller
11: }
```
/home/oherych/go/src/github.com/revel/modules/jobs/app/controllers/status.go line: 25
```
24:	entries := jobs.MainCron.Entries()
25:	return c.Render(entries)
}
```

alkchr:
There is an inconsistency of `runtime.Caller` between gc and gccgo (see `golang/go/issues/6835` for details):

> The stack trace using the gc compiler is
>
> 0: main.WhoCalledMe
> 1: main.(*Inner).meth
> 2: main.main
> 3: runtime.main
>
> Using gccgo it's
>
> 0: main.WhoCalledMe
> 1: main.meth.pN10_main.Inner
> 2: main.meth.pN10_main.Outer
> 3: main.main
> 4: runtime_main

pedromorgan:
Is this a revel bug ??

alkchr:
@pedromorgan Yes and No. On the one hand Revel could address this issue either by using [conditional compilation](https://golang.org/pkg/go/build/#hdr-Build_Constraints) (easy fix), e.g.:
```go
// +build !cgo
```
or by use of code generation I've mentioned above (may require rethinking of the architecture).
On the other hand, this is not really Revel's fault but incompatibility of go compiler with cgo. Did you read [Go Proverbs](https://go-proverbs.github.io/)?

> Cgo is not Go.

Thought, there is also:

> Cgo must always be guarded with build tags.

There is a third option (the easiest), too. Announce that cgo is not supported.

notzippy:
Closing, cgo unsupported at this time

PR #1038, opened 2016-03-29
create the log directory if it does not already exist

chucklqsun:
purpose:
create dir and any necessary parents for log path in app config, such as "log.info.output  = log/%(app.name)s.log"

filemode: 0777


brendensoares:
@chucklqsun why did you close your PR? Is this feature not needed?

jeevatkm:
I think because of incorrect branch target, @chucklqsun submitted #1039 after that.

PR #1039, opened 2016-03-29
create the log directory if it does not already exist

chucklqsun:
creates log's dir with any necessary parents, filemod is 0777
example in app config:
log.info.output  = log/%(app.name)s.log

### DIFF
--- a/revel.go
+++ b/revel.go
@@ -227,6 +227,17 @@ func getLogger(name string) *log.Logger {
 			output = os.DevNull
 		}
 
+		logPath := filepath.Dir(output)
+		if _, err := os.Stat(logPath); err != nil {
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(logPath, 0777); err != nil {
+					log.Fatalln("Failed to create log dir", output, ":", err)
+				}
+			} else {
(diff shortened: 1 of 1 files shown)


revel.go
```diff
@@ -227,6 +227,17 @@ func getLogger(name string) *log.Logger {
 			output = os.DevNull
 		}
 
+		logPath := filepath.Dir(output)
+		if _, err := os.Stat(logPath); err != nil {
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(logPath, 0777); err != nil {
```
jeevatkm: It's better to have `0755` as a FileMode, remaining looks fine.

alkchr: I agree that `0755` is better than `0777`. But even better option is to inherit CHMOD from a parent directory.

jeevatkm:
I have tested this PR locally, it seems it doesn't account for relative vs abs path. For eg.:  config value is `log.info.output = log/%(app.name)s.log`, this PR will create a log directory wherever user runs command  `revel run ...`

So, I will be adding default `revel.BasePath` if it's relative path provided in the log config.
```go
if !filepath.IsAbs(output) {
      output = BasePath + string(filepath.Separator) + output
}
```

Issue #963, opened 2015-07-06
[Feature request] resolve format by extension


zhuharev:
i don't now, how create PR

I suggest add following code to  [ResolveFormat](https://github.com/revel/revel/blob/master/http.go#L72-L92) function before existing `switch`  : 

``` go
import "path/filepath"
...
func ResolveFormat(req *http.Request) string {
    ext := filepath.Ext(req.URL.Path)
    switch ext {
    case ".html":
        return "html"
    case ".json":
        return "json"
    case ".xml":
        return "xml"
    case ".txt":
        return "txt"
    }

    accept := req.Header.Get("accept")

    switch {
    case accept == "",
        strings.HasPrefix(accept, "*/*"), // */
        strings.Contains(accept, "application/xhtml"),
        strings.Contains(accept, "text/html"):
        return "html"
    case strings.Contains(accept, "application/json"),
        strings.Contains(accept, "text/javascript"):
        return "json"
    case strings.Contains(accept, "application/xml"),
        strings.Contains(accept, "text/xml"):
        return "xml"
    case strings.Contains(accept, "text/plain"):
        return "txt"
    }

    return "html"
}
```

browsers usually send `text/html` accept headers

huobazi:
also js as rails

pedromorgan:
https://github.com/revel/revel/issues/1045

Issue #659, opened 2014-07-05
Add extension support to `conf/routes` via `pathtree`


pedromorgan:
I've setup the following routes, attempting to use one handler for all, ie 
/airport.EGLL.html, /airport.EGLL.json and  /airport.EGLL.xml.
made it large for readability
# /airport/:ident.:ext

```
 GET     /airport/:ident.:ext       Airport.AirportHandler
  #GET     /airport/:ident.html         Airport.AirportHtml
  #GET     /airport/:ident.json         Airport.AirportJson
  #GET     /airport/:ident.xml          Airport.AirportXml
```

And my current handler is

```
func (c Airport) AirportHandler(ident, ext string) revel.Result {

    data := navdb.GetAirport(ident) 
    switch ext {
        case "json":
            return c.RenderJson(data)

        case "xml":
            return c.RenderXml(data)
    }
    return c.Render(data)
}
```

However I get a 404 not found..I'm expecting  to assign the part after the dot.

notzippy:
Do this and fix the issue with wildcard mapping being prioritized before other mapping

notzippy:
TODO
 [ ] Fork https://github.com/Jetvp/pathtree/commits/master

notzippy:
This is resolved now . If not please re-open

PR #1048, opened 2016-05-08
support for . extension to return request.Format

u007:
based on rails concept, 
can we apply this?

### DIFF
--- a/http.go
+++ b/http.go
@@ -70,12 +70,11 @@ func ResolveContentType(req *http.Request) string {
 // returning a default of "html" when Accept header cannot be mapped to a
 // value above.
 func ResolveFormat(req *http.Request) string {
-	accept := req.Header.Get("accept")
-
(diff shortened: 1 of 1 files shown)


u007:
travis error?
github.com/bradfitz/gomemcache/memcache
# golang.org/x/sys/unix
../../../golang.org/x/sys/unix/asm.s:8 6a: No such file or directory: textflag.h
github.com/garyburd/redigo/internal
github.com/garyburd/redigo/redis
github.com/robfig/go-cache
The command "go get -v github.com/revel/revel/..." failed and exited with 2 during .
Your build has been stopped.

pedromorgan:
Ok so this was one of wishes from before..

#963 and #659

brendensoares:
#963 has a safer solution, though same end result.

u007:
👍

Issue #1068, opened 2016-06-17
Add ClientIP value into Controller instance


jeevatkm:
Added `ClientIP` field to https://github.com/revel/revel/blob/master/controller.go#L16 populate using [ClientIP](https://godoc.org/github.com/revel/revel#ClientIP)

jeevatkm:
Done!

Fyb3roptik:
This issue was never merged into 0.13.1 What is the planned release for this?

PR #1069, opened 2016-06-18
Added a client ip into controller instance

jeevatkm:
Implementation for #1068

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -26,6 +26,7 @@ type Controller struct {
 	MethodType    *MethodType     // A description of the invoked action type.
 	AppController interface{}     // The controller that was instantiated.
 	Action        string          // The fully qualified action name, e.g. "App.Index"
+	ClientIP      string          // holds IP address of request came from
 
 	Request  *Request
 	Response *Response
--- a/server.go
+++ b/server.go
@@ -51,6 +51,7 @@ func handleInternal(w http.ResponseWriter, r *http.Request, ws *websocket.Conn) 
 	// TODO For now this okay to put logger here for all the requests
 	// However, it's best to have logging handler at server entry level
 	start := time.Now()
+	clientIP := ClientIP(r)
 
 	var (
 		req  = NewRequest(r)
@@ -58,6 +59,7 @@ func handleInternal(w http.ResponseWriter, r *http.Request, ws *websocket.Conn) 
 		c    = NewController(req, resp)
 	)
 	req.Websocket = ws
+	c.ClientIP = clientIP
 
 	Filters[0](c, Filters[1:])
 	if c.Result != nil {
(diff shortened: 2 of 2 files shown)


brendensoares:
We should consider the field name. Revel was inspired from the Rails and Play web frameworks. We can also consider the naming in other popular web frameworks like Node.JS and PHP:

- `request.ip` and `request.remote_ip` [Ruby on Rails](https://stackoverflow.com/questions/10997005/whats-the-difference-between-request-remote-ip-and-request-ip-in-rails)
- `request().remoteAddress()` [Java Play](https://stackoverflow.com/questions/11721215/how-to-get-ip-address-of-the-client-in-play-framework-2-0)
- `request.connection.remoteAddress` [Node.JS](https://stackoverflow.com/questions/8107856/how-to-determine-a-users-ip-address-in-node)
- `Request::$client_ip` and `@_SERVER['REMOTE_ADDR']` [Kohana PHP](https://stackoverflow.com/questions/11565190/kohana-accessing-client-ip)

Does anyone have any preference? I feel there is value in choosing a naming convention that will be familiar to developers who have used other web frameworks.

jeevatkm:
I have read your inputs, below is equivalent to Go `http.Request.RemoteAddr` and it's language level-
- `request().remoteAddress()` [Java Play](https://stackoverflow.com/questions/11721215/how-to-get-ip-address-of-the-client-in-play-framework-2-0)
- `request.connection.remoteAddress` [Node.JS](https://stackoverflow.com/questions/8107856/how-to-determine-a-users-ip-address-in-node)
`@_SERVER['REMOTE_ADDR']` [Kohana PHP](https://stackoverflow.com/questions/11565190/kohana-accessing-client-ip)

Below one, customized by respective framework-
- `request.ip` and `request.remote_ip` [Ruby on Rails](https://stackoverflow.com/questions/10997005/whats-the-difference-between-request-remote-ip-and-request-ip-in-rails)
- `Request::$client_ip` [Kohana PHP](https://stackoverflow.com/questions/11565190/kohana-accessing-client-ip)

Now, let's come to Revel. It can provide -
- `ServerIP` - comes from `app.conf - http.addr` or local `loopback` address
  - Or We call this as `IP` and add documentation explaining this is server ip. Like Ruby SO question
- `ClientIP` or `RemoteIP` in the Controller
  - Both terms is know in the industry and developers familiar to it

jeevatkm:
@brendensoares what is the final take on this?

brendensoares:
Thanks for taking the time to consider our options on this. Let's merge this!

brendensoares:
PS - what can we do to document this new API addition?

jeevatkm:
`ClientIP` method already has godoc, all we need to do is updating controller attributes in the manual.

Issue #273, opened 2013-08-13
PanicFilter trims traces too aggressively


robfig:
Presently, when the panic filter catches a panic, it has a stack like this:

```
/Users/robfig/gocode/src/github.com/robfig/revel/panic.go:19 (0x493a7)
    com/robfig/revel.handleInvocationPanic: ERROR.Print(err, "\n", string(debug.Stack()))
/Users/robfig/gocode/src/github.com/robfig/revel/panic.go:10 (0x5cba9)
    com/robfig/revel.func.039: handleInvocationPanic(c, err)
/Users/robfig/go/src/pkg/runtime/panic.c:229 (0x15271)
    panic: reflect·call(d->fn, (byte*)d->args, d->siz);
/Users/robfig/gocode/src/github.com/robfig/revel/intercept.go:93 (0x5cb3a)
    com/robfig/revel.func.038: panic(err)
/Users/robfig/go/src/pkg/runtime/panic.c:229 (0x15271)
    panic: reflect·call(d->fn, (byte*)d->args, d->siz);
/Users/robfig/go/src/pkg/runtime/panic.c:487 (0x15a13)
    panicstring: runtime·panic(err);
/Users/robfig/go/src/pkg/runtime/os_darwin.c:462 (0x14968)
    sigpanic: runtime·panicstring("invalid memory address or nil pointer dereference");
/Users/robfig/gocode/src/github.com/coopernurse/gorp/gorp.go:949 (0x19328c)
    com/coopernurse/gorp.(*Transaction).Rollback: return t.tx.Rollback()
/Users/robfig/gocode/src/yext/admin/app/controllers/app.go:38 (0x75906)
    Users.PostSpreadsheet: return c.Redirect(routes.Users.Upload())
/Users/robfig/gocode/src/yext/admin/app/controllers/app.go:0 (0x78ee4)
    (*Users).PostSpreadsheet: ???
/Users/robfig/go/src/pkg/reflect/value.go:474 (0x2ffaf)
    Value.call: call(fn, unsafe.Pointer(ptr), uint32(size))
/Users/robfig/go/src/pkg/reflect/value.go:345 (0x2f00b)
    Value.Call: return v.call("Call", in)
/Users/robfig/gocode/src/github.com/robfig/revel/invoker.go:36 (0x48f7f)
    com/robfig/revel.ActionInvoker: resultValue = methodValue.Call(methodArgs)[0]

 ... 15 frames of filters/etc ...
```

[...]

PR #1081, opened 2016-07-10
Less aggressive trimming of stack traces

jchavannes:
Addresses issue #273.


jchavannes:
_Replacing with #1082 which is against the right branch._

PR #1091, opened 2016-08-19
fix FilterController example

voyageth:
fix FilterController example.


jcvernaleo:
This one is pretty minor and looks correct to me.  It probably should just go in.

Issue #1097, opened 2016-09-11
Can't use Session on Production mode


stargt:
The following is a simple controller for setting a session and a cookie.

``` go
func (c App) Index() revel.Result {
  c.Session["foo"] == "bar"
  c.SetCookie(&http.Cookie{Name: "foo2", Value: "Bar2"})
  return c.Render()
}
```

When I run this revel application with `$ revel run myapp dev`, It works fine. I can see the values on the browser's devtools
But, with `$ revel run myapp prod`, there is no related session and cookie values I can see on the devtools, and also some features are not working I have expected. ([Revel's Flash](https://revel.github.io/manual/sessionflash.html) is also unavailable.)

Finally, I figured out that after changing `mode.dev` value from `false`(default) to `true` under `[prod]` section in revel's `app.conf`, It started to work fine.

I couldn't find more details about this problem on [the official manual](https://revel.github.io/manual/index.html).
Is it a bug? Or is there anything I need to check about in this situation before using revel on Production mode?

lujiacn:
I met the same issue today. Everything works fine in dev mode, but in prod mode, flash not work, csrf not work (due to cannot read session).
My previous revel projects no same issue, any thing updated recently?
@jeevatkm

lujiacn:
issue due to revel.go, line 207
```
	CookieSecure = Config.BoolDefault("cookie.secure", !DevMode)
```
The !DevMode should be value false.

jeevatkm:
Implementation was part of [v0.13.0 release](https://github.com/revel/revel/releases/tag/v0.13.0). 
* Defaults to HttpOnly and always secure cookies for non-dev mode #942, PR #943

lujiacn:
go back to this issue.
since "cookie.secure  = false" is defined in app.conf. In revel, should respect config default value.
```
   CookieSecure = Config.BoolDefault("cookie.secure", !DevMode)
```
to
```
   CookieSecure = Config.BoolDefault("cookie.secure", false)
```

shawncatz:
wouldn't we want the default to be true in production? Isn't that the intent of the change?
Disabling this in prod works around the problem, but exposes those apps to potential issues (as mentioned in the referenced issues / prs)

lujiacn:
"A secure cookie has the secure attribute enabled and is only used via HTTPS".
I did not setup https in revel, but run revel behind nginx (SSL) in production mode. cookie.secure = !DevMode will introduce a issue.

shawncatz:
[...]

PR #1102, opened 2016-09-18
changed default CookieSecure to false

lujiacn:
to make session work in prod mode at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 A high productivity, full-stack web framework for the [Go language](http://www.golang.org).
 
-Current Version: 0.13.0 (2016-06-05)
+Current Version: 0.13.1 (2016-06-06)
 
 **As of Revel 0.13.0, Go 1.4+ is required.**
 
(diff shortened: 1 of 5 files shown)


lujiacn:
not sure how to use secure attribute in http.Cookie, but from the example below, for http request, if set secure as true, no cookie set.
So, if do not set CookieSecure, default value is false, not !DevMode, works for http. If set https, may be use CookieSecure "true"
```
package main

import (
        "fmt"
        "net/http"
        "time"
)

func answer(w http.ResponseWriter, r *http.Request) {
        cookie := http.Cookie{Name: "locality", Value: "here",
                Expires:  time.Now().Add(time.Hour),
                HttpOnly: true, MaxAge: 50000,
                Path: "/", Secure: true}
        http.SetCookie(w, &cookie)
        fmt.Fprintf(w, "Yay")
        cooki, err := r.Cookie("locality")
        fmt.Fprint(w, cooki, err)
        //http.Redirect(w, r, "/", http.StatusFound)
}

func main() {
        http.HandleFunc("/", answer)
        err := http.ListenAndServe(":8888", nil)
        if err != nil {
                fmt.Println(err)
        }
}
```

notzippy:
Reference #1097

brendensoares:
@lujiacn we discussed this issue as a team and decided the best solution is to change the default value to be based on the availability of HTTPS that way we avoid any negative side effects with Sessions not working.

The following change should do it:
`CookieSecure = Config.BoolDefault("cookie.secure", HttpSsl)`

Are you able to make this change or should we take control of this PR?

shawncatz:
didn't hear back from original contributor, closing in favor of #1141

PR #1114, opened 2017-01-04
Improve Validation API

torden:
Hi, I Just made validation methods below

- IPAddr : Validation for IP Address format including IPv4 , IPv6 , IPv4-mapped IPv6 with CIDR
- MacAddr : Validation for H/W MAC Address
- Domain : Validation for Domain (Hostname)
- URL : Validation for URL
- PureText : Validation for sanitary string
- FilePath : Validation for File Path

thx

---
This change is [Reviewable](https://reviewable.io/reviews/revel/revel/1114)

### DIFF
--- a/validators.go
+++ b/validators.go
@@ -3,6 +3,7 @@ package revel
 import (
 	"errors"
 	"fmt"
+	"html"
 	"net"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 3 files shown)


validators.go
```diff
@@ -208,3 +214,386 @@ func ValidEmail() Email {
...
+
+/*
+NORMAL BenchmarkRegex-8   	2000000000	         0.24 ns/op
+STRICT BenchmarkLoop-8    	2000000000	         0.01 ns/op
+*/
+const (
+	NORMAL = 0
+	STRICT = 4
+)
+
```
pedromorgan: Can you explain a bit more that NORMAL/STRICT does ?

torden: The difference between them as follow

STRICT : Not allow Invisible characters (control characters without tab, space, newline)
NORMAL : Allow Invisible characters

Welcome back~ @pedromorgan

PR #1129, opened 2017-03-08
Minor fixes to the contribution guide

funyug:
Updated Command:
`git checkout -b feature/useful-new-thing origin/develop`
Current command : 
`git checkout -b feature/useful-new-thing develop`
Problem:
This tries to find the local branch develop which instead should use the remote branch origin/develop.

Updated Command:
`git push fork feature/useful-new-thing`
Current command:
`git push fork develop`
Problem:
This tries to push the changes to the branch develop instead of the feature branch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,8 @@ In order to have your pull requests accepted, we recommend you make your changes
 new git branch. For example,
 ```
 $ git checkout -b feature/useful-new-thing origin/develop    # Create a new branch based on develop and switch to it
-$ ...                                                 # Make your changes and commit them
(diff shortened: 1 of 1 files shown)


notzippy:
Do you think this should be kept in this format, or should we reformat it like this : https://github.com/revel/revel/wiki/Creating-a-Development-Environment

shawncatz:
I think there's a larger project to go through and normalize all of the docs, but for v0.14 we can just push this out

pedromorgan:
Its on the TODO list to explain how to fork/branch in revel, in the manual probably on the qucikref page,

http://revel.github.io/quickref/index.html

shawncatz:
close/reopen to force new build

PR #1136, opened 2017-03-12
Template engine update

notzippy:
This is an update on PR https://github.com/revel/revel/pull/1003 , to make it work with the current development branch. I left the interface the same but modified the main loader somewhat so that the loader would behave better in mixed scenarios which would likely be a lot of the cases 
- Proof of concept app https://github.com/revel/examples/pull/13
- Pongo2 Module https://github.com/revel/modules/pull/22 (this I feel could be better implemented)

Added features
1) Module load order can be specified, this is crucial to ensure the correct views are loaded as desired, ordering is simplistic rather then specifying modules like
```
module.pongo2=github.com/revel/modules/pongo2
module.static=github.com/revel/modules/static
module.jobs=github.com/revel/modules/jobs
```
You now do
```
module.1.pongo2=github.com/revel/modules/pongo2
module.2.static=github.com/revel/modules/static
module.3.jobs=github.com/revel/modules/jobs
```
The stuff between the module and name is dropped when module is loaded

2) Specify engines using single config option `template.engines=pongo2,go` Priority given to first engine on templates but if that fails it will try the secondary and third ones in the list.

3) shebang added to template files, If a template file starts with a shebang, then it is inspected and the engine is used from the shebang line. for example `#! go` on the first line would default to the go template engine to be used to compile that file


runner-mei:
why not use file extension name?

.gohtml -> go template engine
.pghtml - > pongo2 template engine

or

.go.html -> go template engine
.pg.html - > pongo2 template engine

runner-mei:
mixed template is bad idea.

notzippy:
Yes @runner-mei adding extension based switching sounds like a good ideas as well.

pedromorgan:
Changing extention names if a really bad idea imho, hate the idea.

pedromorgan:
Re modules ordering (dont understand why) but can we step in 10's instead,
so we can stick ones in middle, bit like a `run level` kinda thing (eg dovecot config)
eg
```
module.10.pongo2=github.com/revel/modules/pongo2
module.20.static=github.com/revel/modules/static
module.22.my_custom=github.com/me/revel_custom
module.30.jobs=github.com/revel/modules/jobs
```

shawncatz:
I think the filename style (pg.html, etc) is a better choice and matches the way that rails handles preprocessors / templates. 

I would suggest if we're going to do this, that we rearrange them a bit. If a page is html, and uses pongo2, then the filename would be: `index.html.pongo2` (I think the extensions should match the module name)

Thoughts?

shawncatz:
is the load order necessary? Does the config file get interpreted non-deterministically (does the order in the config not correspond to the order in which it's interpreted into a config object)?

notzippy:
@pedromorgan You can step by whatever means you want between modules, as long as it resolves back to something that is sorted analphabetic it will be fine.

@shawncatz I was trying to avoid doing modification to the end of the filename, since a lot of editors are extension type enabled putting in something like `foo.html.pongo2` may not work. @runner-mei suggested an extension format like `foo.pongo2.html` , this would allow editors to still treat the file like an html document and at the same time allow the template loader to use the correct engine to parse it.

@shawncatz It is important to have the modules ordered. Template views are only loaded once by the loader, duplicate top level path names are ignored. Originally the load order was application,revel,random modules. The new load order is application, sorted modules, revel.

notzippy:
@pedromorgan Re extension based naming, yes I didn't like that idea either that is why I have the shebang approach in the code. Having different methods in place should be fine though, we do the same with way controllers operate (we can be specific with what view we want rendered, or we just allow revel to figure it out)

shawncatz:
cool, thanks for the additional context

brendensoares:
Is the current go based templates functionality supported to provide 100% backwards compatibility?

notzippy:
Yes, it is fully backwards compatible, (not case sensitive). I am still tweaking the method signatures a bit yet to streamline some things

notzippy:
I made some modifications to the template engine `Expanded BaseTemplate, to carry template name, template path, template type, and filebytes
Moved template detection into engine, and added a BaseEngine class which detects templates by shebang and file extension`

notzippy:
Closing , updated PR #1157

PR #1141, opened 2017-03-17
changes cookie.secure default to be based on HTTP SSL

shawncatz:
changes cookie.secure default to depend on HTTP SSL setting
should only be enabled when SSL is enabled in revel
also, fixes name of function in comment

### DIFF
--- a/revel.go
+++ b/revel.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/agtorre/gocolorize"
 	"github.com/revel/config"
+	"net/http"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


revel.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	"github.com/agtorre/gocolorize"
 	"github.com/revel/config"
+	"net/http"
```
notzippy: I think this was added by mistake

shawncatz: yeah, probably my IDE trying to be smart, I'll remove

shawncatz: fixed

shawncatz:
@notzippy can you take another look at this, I removed the unintentional additional dep

notzippy:
Can you pull the latest from revel into this PR  @shawncatz ?

shawncatz:
thought I already had, but must have missed something... should be up to date now...

PR #1145, opened 2017-03-18
add go 1.4 back to travis matrix

shawncatz:


pedromorgan:
OK a mojor problem is that 1.6 intorduced the {block} in golang templates..

So i would go with 1.6+ really .. get with it..

PR #1146, opened 2017-03-19
DevMode should be derived from RunMode

shawncatz:


shawncatz:
I need to validate that this actually fixes the original problem (DevMode not set before init() functions) before merging this

notzippy:
Do you have a link to the issue ? I thought the person resolved there own issue by adding a function to `revel.OnAppStart ` which is where things should be initialized. This will not set it any sooner

shawncatz:
yeah, I'm seeing that now... I don't think we can guarantee anything on the app/init.go init() function, because of the way that go initializes things...

Added this to `init()` function in `app/init.go`
```
	revel.OnAppStart(ExampleStartupScript)
	fmt.Printf("DevMode == %t, RunMode == %s\n", revel.DevMode, revel.RunMode)
...
func ExampleStartupScript() {
	// revel.DevMod and revel.RunMode work here
	// Use this script to check for dev mode and set dev/prod startup scripts here!
	revel.INFO.Printf("OnAppStart: DevMode == %t, RunMode == %s\n", revel.DevMode, revel.RunMode)
}
```

Output:
```
INFO  2017/03/19 16:58:18 build.go:190: Cleaning dir tmp
INFO  2017/03/19 16:58:18 build.go:190: Cleaning dir routes
INFO  2017/03/19 16:58:18 build.go:190: Cleaning dir tmp
INFO  2017/03/19 16:58:18 build.go:190: Cleaning dir routes
**DevMode == false, RunMode ==**
INFO  2017/03/19 16:58:19 revel.go:365: Loaded module testrunner
INFO  2017/03/19 16:58:19 revel.go:365: Loaded module static
INFO  2017/03/19 16:58:19 revel.go:230: Initialized Revel v0.14.0-dev (TBD) for >= go1.4
INFO  2017/03/19 16:58:19 main.go:30: Running revel server
Go to /@tests to run the tests.
INFO  2017/03/19 16:58:19 init.go:58: **OnAppStart: DevMode == true, RunMode == dev**
```

shawncatz:
I'm actually going to reopen this, because I think it's better to have DevMode be derived from RunMode

shawncatz:
so this change is just changing the way in DevMode is set... it is now derived from RunMode == "dev", instead of a hidden default value in the config.

shawncatz:
nevermind, `mode.dev` is used to override run mode in application config... so this will need a bit more consideration.

I think we *should* change it to just specify the mode directly (mode = dev, mode = prod), but we don't need to do that as part of v0.14

PR #1150, opened 2017-03-24
v0.14.0

brendensoares:


shawncatz:
all of these changes were reviewed before, so I don't think we need to review in detail

PR #1153, opened 2017-03-28
Prevent XSS

ptman:
Revel has an XSS vulnerability that I've fixed this way. There are many ways to fix it, and one should be chosen and used. Another option would be to do something like https://github.com/roblillack/mars/issues/1 and https://github.com/roblillack/mars/issues/2


notzippy:
Thanks for the heads up and test case. We will prioritize this issue.

Issue #1095, opened 2016-08-29
Request Params: Query takes precedance over path


pedromorgan:
Here's the route

```
GET /ajax/ags/abbrev/:code     Ags.Abbrev
```

and controller

``` go
func (c Ags) Abbrev(code string) revel.Result {
    fmt.Println("arg=", code)
    fmt.Println("query=", c.Request.URL.Query().Get("code"))
}
```

and the url + output

```
http://localhost:9000/ajax/ags/abbrev/CMPG_TYPE?code=CMPG_TYPEsss

GET /ajax/ags/abbrevs
arg= CMPG_TYPEsss
query= CMPG_TYPEsss

```

Whats scared me is the somehow the `?query`  is overriding the "path" extraction.

from here its says their "all in one params" http://revel.github.io/manual/parameters.html
which is fair, but is this not dangerous ??

The expected behaviour  would be that the "path" part would override the "query part", or am i missing something ??

pedromorgan:
So digging down for curious.. Think its here calcValues/.. Should be called "parseValues"
https://github.com/revel/revel/blob/master/params.go#L75

and probalby here is clash.. as we expect the "path" to be set after ALL query stuff etc..
https://github.com/revel/revel/blob/master/params.go#L83

Anyway, they are sepeate things.. after looking at code..
But we HAVE to make this clear

pedromorgan:
```csv
## Hot Zone
GET /reveldevil/params                        RevelDevil.ParaHell
GET /reveldevil/params/:code                  RevelDevil.ParaHell
GET /reveldevil/params/:code/:hell            RevelDevil.ParaHell
GET /reveldevil/params/:code/:hell/:haven     RevelDevil.ParaHell
```

and the controllers
```
func (c RevelDevil) ParaHell(code string) revel.Result {

	stuff := make(map[string]interface{})
	stuff["0_url="] =  c.Request.URL.String()
	stuff["1_aqquiring a param named="] = "code"
	stuff["2_revel_params="] = c.Params
	stuff["3_controller_method_arg="] = code
	stuff["4_revel_values="] = c.Params.Values.Get("code")
	stuff["5_raw_query_arg"] = c.Request.URL.Query().Get("code")
	stuff["6_revel_route/:code"] = c.Params.Route.Get("code")
	stuff["7_revel_query?code="] = c.Params.Query.Get("code")


	return c.RenderJson(stuff)
}
```
results =
```js
{
[...]

Issue #503, opened 2014-02-13
Access to any project files using Static module


mcheng89:
By passing the prefix param in the url, I am able to access code under the application directory.

Examples: 
/tmp/main.go?prefix=app/
/routes?prefix=conf/

pedromorgan:
ok @alkchr I wanna stick your above post in the manual  http://revel.github.io/manual/parameters.html

is that ok. .snag for me is the pages of app.conf and paramaters need to be on same page .. maybe

PR #1155, opened 2017-03-31
param map overwrite priority

0xrofi:
This pull-req fix the problem that Fixed parameters can be overwritten with Get parameters.

Here is the reproduction scenario, `routes`

```
GET     /hoge            App.Hoge("fixed")
```

and controller

```go
func (c App) Hoge(param string) revel.Result {
	fmt.Println(param)
	return c.Render()
}
```

and url

```
http://localhost:9000/hoge?param=get
```

exptected output: `fixed`
but actual: `get`

Issue #1095 and pr #1126 change the order of params priority, but insufficient.
So this pr is overwrite low priority params by higher priority params.

### DIFF
--- a/params.go
+++ b/params.go
@@ -107,15 +107,15 @@ func (p *Params) calcValues() url.Values {
 	}
 	// form vars overwrite
 	for k, v := range p.Form {
-		values[k] = append(values[k], v...)
+		values[k] = v
 	}
 	// :/path vars overwrite
 	for k, v := range p.Route {
-		values[k] = append(values[k], v...)
+		values[k] = v
 	}
 	// fixed vars overwrite
 	for k, v := range p.Fixed {
-		values[k] = append(values[k], v...)
+		values[k] = v
 	}
 
 	return values


notzippy:
Can you update params_test.go with a test case for this situation ? thanks..

pedromorgan:
Confused.. The `fixed` should overwrite all other params, but all other params need to be arrays ?

0xrofi:
@notzippy This situation is not only params but also include route, binder, controller. So I don't know how to write this situation's test case... sorry.

@pedromorgan Because first element of array is used to controller's arguments. So just like as I said above, `Route` parameters and `Form` parameters also can be overwritten with `Query` parameters. (Sorry for `Get` in I said above, it's a mistake of `Query`)
[problem of #1095](https://github.com/revel/revel/issues/1095#issue-173892886) isn't fixed yet and I think this patch fixes it.

If all params(`Query`, `Form`, `Route` and `Fixed`) need to append to array, I think that it should change the order of appending parameters (it's discussed in #503 ).

notzippy:
Would this be better ? So highest gets the first slot but lower ones can be promoted...
```
	for k, v := range p.Fixed {
-		values[k] = append(values[k], v...)
+		values[k] = append(v, values[k]...)
 	}

**Comment by 0xrofi** [comment]
Alternatively I suggest here instead of this PR.

```diff
        // Copy everything into a param map,
-       // order of priority is least to most trusted
+       // order of priority is most to least trusted
        values := make(url.Values, numParams)

-       // ?query vars first
-       for k, v := range p.Query {
+       // fixed vars first
+       for k, v := range p.Fixed {
                values[k] = append(values[k], v...)
        }
-       // form vars overwrite
-       for k, v := range p.Form {
+       // :/path vars append
+       for k, v := range p.Route {
                values[k] = append(values[k], v...)
        }
-       // :/path vars overwrite
-       for k, v := range p.Route {
+       // form vars append
+       for k, v := range p.Form {
                values[k] = append(values[k], v...)
        }
-       // fixed vars overwrite
-       for k, v := range p.Fixed {
+       // ?query vars append
+       for k, v := range p.Query {
                values[k] = append(values[k], v...)
        }
```

notzippy:
What about a hybrid approach ? Since fixed and route parameters should never be set using query or form values they always do a replace, and query and form parameters are appended together.
```
	// ?query vars are first
	for k, v := range p.Query {
		values[k] = append(values[k], v...)
	}
    
	// form vars append
	for k, v := range p.Form {
		values[k] = append(values[k], v...)
	}
    
	// :/path vars overwrite
	for k, v := range p.Route {
		values[k] = v 
	}
    
	// fixed vars overwrite
	for k, v := range p.Fixed {
		values[k] = v 
	}
```

pedromorgan:
@notzippy Yes that is what we want.. All query/post etc vars appended.. BUT a fixed one is CONST ;-) ta.. commit it bro...

pedromorgan:
Need to update manual to explain this problem.. and also a new CORS chapter.. ;-)

notzippy:
@pedromorgan I think with this PR #1157 you may need to rewrite the manual :-P
I incorporated this change in there..

brendensoares:
Closing in favor of the above defined hybrid approach to be released with PR #1171. Thanks for everyone's effort on this!

PR #1161, opened 2017-04-17
receive a post method of json request.

yangfei4913438:


controller.go
```diff
@@ -135,6 +137,22 @@ func (c *Controller) RenderTemplate(templatePath string) Result {
 	}
 }
 
+//receive a post method of json request.
```
xiaozi0lei: // ReceiveJSON unmarshal the input streaming JSON data, returns `map[string]interface{}` type data.
// To access this data we can use a type assertion to access map[string]interface{}. 
// Reference to [Decoding arbitrary data chapter] in https://blog.golang.org/json-and-go

xiaozi0lei: I think a little clear. :)

notzippy:
Looks like a good start - It would be nice to try and implement a more transparent solution but that can be deferred.

brendensoares:
Covered by #1173

PR #1162, opened 2017-04-24
Adds  path detection when loading modules

brendensoares:
Need to consider the side effects. At the very least, this allows me to run my module's tests along side the main app's tests.

Also, we should refactor the "hack"; is there a better way to enables testing in an app other than a hard coded module check like this?

Is there a better way to enable tests per module?


notzippy:
Recommend passing in a flag from the revel/cmd to indicate the core app that revel is to add in all folders for modules that contain tests. Additional feature could be to filter tests by module

brendensoares:
@notzippy I kind of prefer doing that via `app.conf` if at all. I think our goal should to use convention over configuration.

It makes sense that if the module has a `tests` folder that we make use of it. Are there scenarios where that would backfire or be negative?

notzippy:
I was just thinking there is no need to include those folders if revel was not running any tests - so harness would pass in something indicating what mode it should be run in..

PR #1180, opened 2017-06-03
Add template func isEmpty

Acidic9:
Added a new function to revel `IsEmpty(v reflect.Value) bool` and included it as a template function `isEmpty` which can be used to check if an interface's value is zero'ed.

An empty string, 0, empty struct, etc. will return true when parsed as the argument to `IsEmpty`.

**An example of how it can be used in a template:**
  `You are {{if (isEmpty .Username)}}not logged in{{else}}{{.Username}}{{end}}!`
Which would result in either
  `You are not logged in!`
or
  `You are Acidic9!`

### DIFF
--- a/template.go
+++ b/template.go
@@ -63,6 +63,9 @@ var (
 			}
 			return template.JS("")
 		},
+		"isEmpty": func(v interface{}) bool {
+			return IsEmpty(reflect.ValueOf(v))
+		},
 		"field": NewField,
 		"firstof": func(args ...interface{}) interface{} {
 			for _, val := range args {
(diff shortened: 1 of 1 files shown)


notzippy:
Hi, that looks like a great addition. We are just about ready to release a configurable template engine. Which will affect this PR. I will ping you when it is released so you can update this code

Acidic9:
Only one day ago I discovered what {{with .}} actually does and I'm not too sure if my isEmpty func is as necessary. It's fine if it doesn't get merged because it may not be a necessity.
Thanks though.

notzippy:
@Acidic9 Did you want to resolve these conflicts ? Or should I just close because the `{{with` does the same ?

notzippy:
Closing, using `{{with` accomplishes the same because the inner block will only be executed if object is not nil

PR #1185, opened 2017-06-14
Fixes wildcards in router files in app from affecting modules or visa…

notzippy:
## Issue
The route configuration file could affect packages outside its scope when doing wildcard searches. For example for the booking application has the following route

    *       /:controller/:action                  :controller.:action

So requests to `http://localhost:9000/TestRunner/Index` would invoke the `http://localhost:9000/@test` page even though the module testsuite is not configured to allow the wildcard matches. This patch ensures that a wildcard match made for a controller must be specified in the controllers local route file. ie the Application route file cannot wildcard match a controller in another module.

Also update the 404 template to show routes with their module name

## Namespace 

Namespace definition begins with the application configuration. Modules are defined inside the configuration file by `module.static=github.com/revel/modules/static` this defines the revel namespace of the word `static` to all the controllers and templates within the module. To reference the name space use the backslash in front of your controller action like below.

Route mapping inside the route file would be defined like

    *       /favicon.png                  static\Static.Serve("folder")

Module routing remains unchanged

    *      /debug/                          module:testrunner

Reverse routing has an optional namespace as well.

    <td><a href="{{url `jobs\Jobs.Run` $index}}">Start</a></td>

#### Referencing Namespace within the route file. 
Given you had a line in the configuration importing a module like   `module.jobs=jobs` Route mapping inside the route file would be defined like.

    *       /run/:id                  jobs\Jobs.Run

To remain backwards compatible routes without a namespace will be matched to the first controller/name within the same module space or global space. Another words, if you have a definition like below in your app folder it will be matched to the first app controller that is called `Jobs` with a function `Run`

    *       /run/:id                  Jobs.Run

#### Wildcard matching
You can limit the wildcard matching to a single module by having the namespace in the route like the following. 

    *       /jobs/:controller/:action                  jobs\:controller.:action

Any route definitions in a module will be assumed that they are limited to only that module, unless that module prefixes a route path with a namespace 

#### Referencing Namespace within the template. 
You can reference a controller by its namespace within the application template by adding the `namespace\\` operator before the controller action. For example like given the module was defined in `app.conf` like  `module.myjobs=jobs` then to reference the controller action Jobs.Run the namespace should be appended in front like

    <td><a href="{{url `myjobs\Jobs.Run` $index}}">Start</a></td>

To remain backwards compatible the namespace reference is not required and the synatx `{{url \`Jobs.Run\` $index}}` will work, but it will use the first controller called `Jobs` with the action `Run` to perform the reverse route.

### Namespaces in a Module 
When building a module the application namespace that the module is imported into is unknown. So we introduce a special letter sequence **\_LOCAL\_\\** to indicate that these letters will need to be replaced within the template before the engine receives it. Then the template reverse routes function (the `url` template function) you should define the key `_LOCAL_(.module)\` in front of `url`. The `(.module)` portion is only needed if this is referencing a different module in another namespace (see next section for more details) . When revel reads the template file it will rewrite the `_LOCAL_(.module)\` portion with the namespace of the defined for example:

    <td><a href="{{url `_RNS_\Jobs.Run` $index}}">Start</a></td>

Given that the module is imported as `module.myjobs=jobs` The template would then be realized as 

    <td><a href="{{url `myjobs\Jobs.Run` $index}}">Start</a></td>

**Note** to remain backwards compatible the revel engine will attempt to reverse lookup a `Controller.Action` across all available controllers. If two controllers have the same name and action the default will be to match the first one. This will be slightly slower then a fully qualified lookup (one that has a namespace defined) is.

If a module requires another module, then there may be a conflict between how that module imported its module vs how the application imported the module. The correct way to map a imported module would be to specify the module to be included in the app.conf of the module like `module.myjobs=jobs`, then reference it in the template like the following

    <td><a href="{{url `_LOCAL_.myjobs\Jobs.Run` $index}}">Start</a></td>

In a module routes file this would be similarly done like

    *       /favicon.png                  _LOCAL_.static\Static.Serve("folder")


router_test.go
```diff
@@ -133,7 +133,7 @@ var routeMatchTestCases = map[*http.Request]*RouteMatch{
 		Method: "GET",
 		URL:    &url.URL{Path: "/"},
 	}: {
-		ControllerName: "Application",
+		ControllerName: "application",
```
brendensoares: Why did the case of these change?

namespace.go
```diff
@@ -0,0 +1,37 @@
+package revel
+
+import (
+	"regexp"
+	"bytes"
+)
+
+// Module matching template syntax allows for modules to replace this text with the name of the module declated on import
```
brendensoares: `declated`

controller.go
```diff
@@ -40,6 +40,9 @@ type Controller struct {
 	Validation *Validation            // Data validation helpers
 }
 
+// The map of controllers, controllers are mapped by using the namespace|controller_name as the key
```
brendensoares: should godocs start with the name of the identifier followed by a description? E.g. "controllers is a list of controllers mapped using..."?

template.go
```diff
@@ -184,8 +182,8 @@ func (loader *TemplateLoader) Refresh() (err *Error) {
 	return loader.compileError
 }
 
-// WatchDir returns true of directory doesn't start with . (dot)
-// otherwise false
+// Checks to see if template exists in templatePaths, if so it is skipped (templates are imported in order
+// reads the template file into memory, replaces namespace keys with module (if found
```
brendensoares: word choice "in order reads the template" feels awkward

template.go
```diff
@@ -205,6 +203,12 @@ func (loader *TemplateLoader) findAndAddTemplate(path, fullSrcDir, basePath stri
 		ERROR.Println("Failed reading file:", path)
 		return
 	}
+	// Parse template file and replace the "_RNS_|" in the template with the module name
```
brendensoares: Better symbol other than `RNS` e.g. `LOCAL`?

template.go
```diff
@@ -205,6 +203,12 @@ func (loader *TemplateLoader) findAndAddTemplate(path, fullSrcDir, basePath stri
...
+	// allow for namespaces to be renamed "_RNS_(.*?)|"
```
brendensoares: namespace vs module name?

template_functions.go
```diff
@@ -0,0 +1,244 @@
...
+	"html/template"
+	"reflect"
+	"strings"
+	"time"
+	"bytes"
+	"errors"
+)
+
+var (
+	// The functions available for use in the templates.
```
brendensoares: godoc starts with identifier name?

router_test.go
notzippy: The case is the normalized to lowercase. Just to make all lookups consistent.

namespace.go
```diff
@@ -0,0 +1,37 @@
...
+// this allows the reverse router to use correct syntax
+// Match _RNS_.static or  _RNS_|
```
Fyb3roptik: Should we do \ (Backslash) instead of a | (pipe)?

PR #1191, opened 2017-07-04
reduce the number of calls to findTarget

namusyaka:


intercept.go
```diff
@@ -66,14 +66,15 @@ type Interception struct {
 	interceptAll bool
 }
 
+type interceptionEntry struct {
+	intc *Interception
+	arg  reflect.Value
```
brendensoares: `target` instead of `arg`?

intercept.go
```diff
@@ -161,15 +162,16 @@ func InterceptMethod(intc InterceptorMethod, when When) {
 	})
 }
 
-func getInterceptors(when When, val reflect.Value) []*Interception {
-	result := []*Interception{}
+func getInterceptorEntries(when When, val reflect.Value) []*interceptionEntry {
+	result := []*interceptionEntry{}
```
brendensoares: It's a small and simple optimization to reduce the calls to `findTarget()`, but is it possible to cache these entries on app start for further performance gains?

namusyaka: The controller (and `AppController`) value will be initialized for each request now, so that we should require large modifications to improve the further performance, I think.

intercept.go
namusyaka: Thanks. Done https://github.com/revel/revel/pull/1191/commits/6fdf79db19ccbe2e58e952a841e3dae056d4dec0

notzippy:
Would you be able to resolve the conflicts against the develop branch ? 
thanks

notzippy:
Closing, the better method to use is the Before/after instead of the interceptor.

PR #1207, opened 2017-07-22
Timeago Template Func

wariosolis:
Added new template function (Go) called "timeago" using https://github.com/xeonx/timeago

Documentation added to the website.
If the PR is accepted, consider the documentation https://github.com/revel/revel.github.io/pull/133.

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	//"github.com/revel/config"
 	"github.com/xeonx/timeago"
 	"html"
 	"html/template"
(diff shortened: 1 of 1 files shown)


template_functions.go
```diff
@@ -145,6 +146,35 @@ var (
 		},
 		"slug": Slug,
 		"even": func(a int) bool { return (a % 2) == 0 },
+
+		// Using https://github.com/xeonx/timeago
+		"timeago": func(viewArgs map[string]interface{}, datetime time.Time) string {
```
notzippy: If possible can you modify this so it behaves similar to the i18template function. 
Basically allow for variable number of arguments
- If one argument format the time as english
- if two arguments, and second one is string assume that is the region
- if two arguments, and second is map assume that is the viewArgs and extract the CurrentLocaleViewArg from the map

template_functions.go
```diff
@@ -145,6 +146,35 @@ var (
...
+
+			localeStr, ok := viewArgs[CurrentLocaleViewArg].(string)
+			if !ok || strings.Contains(localeStr, "en") {
+				return timeago.English.Format(datetime)
+			}
+			// internalization by current locale
+			var customLanguage = timeago.Config{
```
notzippy: Store the customLanguage outside the function in a map keyed by the localeStr, so that a new timeago.Config object doesnt get created every time a this template function is called

wariosolis:
@notzippy Please check the changes, I'll wait for your comments.

template_functions.go
```diff
@@ -242,3 +248,77 @@ func Slug(text string) string {
...
+
+func TimeAgo(args ...interface{}) string {
+
+	datetime := time.Now()
+	lang := ""
+	var viewArgs interface{}
+	switch len(args) {
+	case 0:
+		ERROR.Printf("No arguements passed to template call")
+	case 1:
```
notzippy: Probably shoud say no arguements passed to TimeAgo call

PR #1211, opened 2017-07-25
Before, after, panic, finally controller auto rule implementation

### LINKED ISSUE #420: Controller Before() / After()
Would be nice if all controllers had before and after methods that ran automatically if present.

notzippy:
### Description
Smart method application - controllers that have methods named `Before`,`After`,`Finally`,`Panic`, will be called in the same manner as the interceptor does. The method signature is a little different then normal to allow for method Overriding and cascades. ie if **Application** has **GorpController** controller embedded inside it and both define the `Before` method then what happens is the `GorpController.Before` is invoked before the `Application.Before`. The method signature must match the following syntax though 

```
func (c Application) Before() (r revel.Result, a Application) {
...
func (c *GorpController) Before() (result revel.Result, controller *GorpController) {
```
So notice along with the revel.Result we return the Controller type ? This is so when the application starts up it can determine which method belongs to what field in the controller (currently Go does not tell you this information). So other then that little change in signatures these work identical to the revel.Interceptors without any configuration.

The mapping to the function is done on load so firing these functions is very fast. On the `Before` method the deepest embedded struct with the `Before` method is run first, then the next deepest and so on until the top level. `After`, `Finally`, `Panic` are run from the top down

Tested with the booking app (removed interceptors for the GorpController and renamed methods to match) worked like a charm

### Implementation
Add the filter `revel.BeforeAfterFilter` to revel.Filters on app/init.go 

### Breaking Changes
There should be none, on startup it does a deep search through the controllers to match the method name when the name matches it also checks that the method has 2 arguments and the second argument must be the same type as the struct the method is defined for.

closes #420

### DIFF
--- a/before_after_filter.go
+++ b/before_after_filter.go
@@ -1,0 +1,60 @@
+package revel
+
+import (
+	"reflect"
+)
+
+// Autocalls any defined before and after methods on the target controller
+// If either calls returns a value then the result is returned
+func BeforeAfterFilter(c *Controller, fc []Filter) {
+	defer func() {
+		if resultValue := beforeAfterFilterInvoke(FINALLY, c); resultValue != nil && !resultValue.IsNil() {
+			c.Result = resultValue.Interface().(Result)
+		}
+	}()
+	defer func() {
+		if err := recover(); err != nil {
+			if resultValue := beforeAfterFilterInvoke(PANIC, c); resultValue != nil && !resultValue.IsNil() {
+				c.Result = resultValue.Interface().(Result)
+			}
+			panic(err)
+		}
+	}()
+	if resultValue := beforeAfterFilterInvoke(BEFORE, c); resultValue != nil && !resultValue.IsNil() {
+		c.Result = resultValue.Interface().(Result)
+	}
+	fc[0](c, fc[1:])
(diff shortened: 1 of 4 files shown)


before_after_filter.go
```diff
@@ -0,0 +1,60 @@
...
+	case FINALLY:
+		index = c.Type.ControllerEvents.Finally
+	case PANIC:
+		index = c.Type.ControllerEvents.Panic
+	}
+
+	if len(index)==0 {
+		return
+	}
+	for _,function:= range index {
```
brendensoares: ` :=` space

brendensoares: `_, f` space

before_after_filter.go
```diff
@@ -0,0 +1,60 @@
...
+	if resultValue := baInvoke(BEFORE, c); resultValue!=nil && !resultValue.IsNil() {
+		c.Result = resultValue.Interface().(Result)
+	}
+	fc[0](c, fc[1:])
+	if resultValue := baInvoke(AFTER, c); resultValue!=nil && !resultValue.IsNil()  {
+		c.Result = resultValue.Interface().(Result)
+	}
+}
+
+func baInvoke(method When, c *Controller) (r *reflect.Value) {
```
brendensoares: `ba` means what?

brendensoares: Before After...is there a better name or meh?

PR #1213, opened 2017-07-26
Logger update

notzippy:
This logger provides context logging for the controller, and it provides the ability to output in either json or text file format. Log messages and key value pairs. File rotation is also handled based on either the file size or the number of days specified in the app.config

# Logging
By default a new logger is available for the application called revel.AppLog - its interface is using the log15 interface (see [here](https://github.com/inconshreveable/log15)) You can assign log handlers which can perform multiple functions. 

**NOTE** The new logger expects message values in key value pairs. For example `revel.AppLog.Debug("Hi there")` is fine `revel.AppLog.Debug("Hi there", 25)` will panic (only passed one argument for the context). `revel.AppLog.Debug("Hi there","age",25,"sex","yes","state",254)` is fine. and will produce a log message that includes the context age,sex,state.

The logger in the controller adds the ability to extend its context so you can say `c.Log = c.Log.New("user",userid)` and all subsequent logs using that controller will report the `'user'=userid` in the log files. Each request instance in the controller will create a new log context so all your logs messages during the request will be unique. Revel create the `c.Log` using a context that includes the ip, request path and request method

### Default operation
By default it operates the same as the old logging utility. Unless the log file ends in `json` , in that case the log output will be in json format. Log file size is limited to 10G in size but may be specified in app.conf `log.maxsize` (in MB). Log file age is limited to 14 days be default unless specified in app.conf `log.maxage` (in days). If either is exceeded the file is date stamped and a new file is created.

# Deprecated loggers
`revel.ERROR`,`revel.WARN`,`revel.INFO`,`revel.TRACE`, are deprecated. Controllers in applications should log to the context logger `controller.Log` and jobs should log to `revel.AppLog`

Sample Screen Shot
image

### TODO

- [x] Go through revel remove all references to deprecated loggers
- [x] docs for new `app.conf`


logger.go
```diff
@@ -0,0 +1,394 @@
...
+	termTimeFormat = "2006/01/02 15:04:05"
+	floatFormat    = 'f'
+	termMsgJust    = 40
+	errorKey       = "LOG15_ERROR"
+)
+
+// INFO  2017/07/25 14:53:40 db.go:768: [gorp] begin; [] (138.433µs)
+func terminalFormat(noColor bool) log15.Format {
+	return log15.FormatFunc(func(r *log15.Record) []byte {
+		var color = 0
```
brendensoares: add comments for what colors these correspond to

logger.go
```diff
@@ -0,0 +1,394 @@
...
+			k, ok := r.Ctx[i].(string)
+			if k == "caller" || k == "fn" {
+				continue
+			}
+			v := formatLogfmtValue(r.Ctx[i+1])
+			if !ok {
+				k, v = errorKey, formatLogfmtValue(k)
+			}
+
+			// XXX: we should probably check that all of your key bytes aren't invalid
```
brendensoares: `XXX`? Maybe `TODO`

logger.go
```diff
@@ -0,0 +1,394 @@
...
+		case log15.LvlInfo:
+			color = 32
+		case log15.LvlDebug:
+			color = 36
+		}
+
+		b := &bytes.Buffer{}
+		lvl := strings.ToUpper(r.Lvl.String())
+		if noColor == false && color > 0 {
+			// fn := findInContext("fn",r.Ctx)
```
brendensoares: delete?

logger.go
```diff
@@ -0,0 +1,394 @@
...
+			caller := findInContext("caller", r.Ctx)
+
+			fmt.Fprintf(b, "\x1b[%dm%s\x1b[0m %s %s: %s ", color, toRevel[r.Lvl], r.Time.Format(termTimeFormat), caller, r.Msg)
+
+		} else {
+			fmt.Fprintf(b, "[%s] [%s] %s ", lvl, r.Time.Format(termTimeFormat), r.Msg)
+		}
+
+		// try to justify the log output for short messages
```
brendensoares: `Try`

logger.go
```diff
@@ -0,0 +1,394 @@
...
+
+}
+
+// This structure and method will handle the old output format and log it to the new format
+type loggerRewrite struct {
+	Logger MultiLogger
+	Level  log15.Lvl
+}
+
+func (s loggerRewrite) Write(p []byte) (n int, err error) {
```
brendensoares: `s` vs `lr`?

server-engine.go
```diff
@@ -154,13 +156,10 @@ func handleInternal(ctx ServerContext) {
 	// RequestStartTime ClientIP ResponseStatus RequestLatency HTTPMethod URLPath
 	// Sample format:
 	// 2016/05/25 17:46:37.112 127.0.0.1 200  270.157µs GET /
```
brendensoares: Update the comment

notzippy:
@brendensoares Did a bigger update to the code base - I moved most of the logger code inside a `revel/logger ` package and broke out the code into separate files. There still is a revel/logger.go file which defines the global loggers but in general it is cleaner. Can you have a quick peek again ?

zuoRambo:
I think `revel.TRACE` `revel.INFO` `revel.WARN` `revel.ERROR` should keep.
And seperate controller log and job log is a good idea! 👍

notzippy:
@zuoRambo thanks for the feedback - The reason we choose to deprecate those variables is because the functionality is contained inside the revel.AppLog, `revel.AppLog.Error("message")` `revel.AppLog.Warn("message")` `revel.AppLog.Debug("message")` `revel.AppLog.Info("message")` `revel.AppLog.Crit("message")`. Since they are contained there we should not expose a duplicate method elsewhere.

I realize it may be some work to switch over existing applications to use the new logging, but with the new logging you get some extended features like JSON log records with field/values. 

Let me know if you have more reasons for revel.ERROR log messages to remain open

PR #1214, opened 2017-07-26
Delete repeated comment

eXMooR:


pedromorgan:
>> It's awkward 

Should that be `akward`

PR #1216, opened 2017-07-27
Session engine support.

notzippy:
Configurable session engine, register interface revel.SessionRegisterSessionEngine(f func() revel.SessionEngine, name string)
Session engine is chosen from app.conf value session.engine
Currently only engine type is revel-cookie.
Adds the feature to store interface{} session from controller using SessionMarshal, SessionUnmarshal, SessionDelete methods.


notzippy:
I am not really sure this will work in all situations

pedromorgan:
Look good to me, I;m already serialising json, so this solves my problem..

notzippy:
@pedromorgan 
I am torn if I should be adding two new methods (and two maps) on the controller to do this. Or if this should be implemented elsewhere and injected onto the controller as a single object - my other thoughts are on `page session` data (data that would be persisted through a series of JSON requests, ie only be relevant for that page)

notzippy:
I am testing a new approach where the c.Session object is changed to a `map[string]interface{}` Since this is a breaking change I am deferring this PR till next release...

Issue #1188, opened 2017-06-19
Fatal error: concurrent map writes - goroutine


alexfromvl:
I use testing of the created site through WRK:

`wrk -t1 -c3 -d10S --latency "http://127.0.0.1/"`

And every time after a few seconds after starting the test I get an error:

```
fatal error: concurrent map read and map write

goroutine 27677 [running]:
runtime.throw(0x8a54ce, 0x21) 
...

goroutine 1 [IO wait]:
net.runtime_pollWait(0x7f48b9b00f00, 0x72, 0x0)
...

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
...

goroutine 28027 [IO wait]:
net.runtime_pollWait(0x7f48b9b00e40, 0x72, 0x7)
...

```

Is it possible to avoid this error during testing?

notzippy:
Sorry, this had gotten buried in my email. Is this an app that you have written or are you using one of the example apps ?

alexfromvl:
Hello. The problem arises when using the example [https://revel.github.io/tutorial/firstapp.html](https://revel.github.io/tutorial/firstapp.html) and when creating a simple site from one main page.

By this error I was able to find this discussion[https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations](https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations). But it's not entirely clear how to rewrite the code so that you do not come to this error?

daxorid:
I have reproduced this, on a production site with normal traffic. 

The templateLoad() function in template.go:332 is extremely concurrency-unfriendly.

I have hacked up a workaround for this by wrapping all accesses of loader.TemplateMap inside Lock/Unlock blocks until such time as the revel people address this.

Hope this helps.

notzippy:
Thanks, PR #1205 should resolve this issue

runner-mei:
PR #1205 is not resolve this issue

notzippy:
Thanks @runner-mei , I had thought that maps with multiple read access was OK and writes required synchronized access. It appears I would need to use a `sync.RWMutex` for reading and writing to the map. I will update this PR with that which should make it functionally the same as yours.

notzippy:
@runner-mei I have created a new PR to address the issue further - I think there may have also been a basic flaw with the previous PR where the save map values were never used. Please checkout #1223

runner-mei:
[...]

PR #1223, opened 2017-08-17
Fix for the template map concurrency issue

notzippy:
Another update, to address the concurrency issue with the template map. Changed the templatemap to an atomic value that is only read from. New items are added to a copy of the template map after locking using a sync.Mutex lock . Then the template map is replaced / updated. As recommended by @runner-mei here #1188


notzippy:
Closing in favor of #1225

PR #1224, opened 2017-08-18
Feature/graceful shutdown

zuoRambo:
import and use `github.com/facebookgo/grace` to handle gracefully shutdown ~


zuoRambo:
We could upgrade `go` version to *1.8*, so we can use `Shutdown` and `context`.
```go
var wg sync.WaitGroup
signalChan := make(chan os.Signal, 1)
signal.Notify(signalChan, syscall.SIGINT, syscall.SIGTERM)

go func() {
	<-signalChan
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(Config.IntDefault("app.cancel.timeout", 10)))
	defer cancel()
	Server.Shutdown(ctx)
	runShutdownHooks(&wg)
}()
```

notzippy:
I was thinking of just adding another configurable port to listen on so you could shutdown the server or even get some running stats from the server

notzippy:
@zuoRambo , we have merged code into develop that allows you to define any http server. See [fasthttp](https://github.com/revel/modules/tree/develop/server-engine/fasthttp) for an example implementation. I think this would be an ideal implementation for the graceengine idea. The standard go server source is [here](https://github.com/revel/revel/blob/develop/server_adapter_go.go). Can you re-implement this PR as a server-engine module (modules/server-engine/grace)?

zuoRambo:
OK，thanks @notzippy 
For graceful reload our application, we should catch the terminated signal and notify the server to finish the established request with timeout option, and notify our application which have some customized asynchronous jobs.

zuoRambo:
@notzippy I merge the latest master branch

notzippy:
@zuoRambo Thanks a lot, the implementation looks great... Unfortunate that this feature is only available in go 1.8. A lot of peoples installations are on Ubuntu which seems to be stuck on Go 1.6 (which is why we kept compatibility to that version.) 

Perhaps this can implement this by making a new server engine wrapping https://github.com/braintree/manners  ? Which is compatible with Go 1.6 and supports the shutdown method. What are your thoughts on this ?

jansila:
hi guys, 

any progress on this? :)

notzippy:
I have been looking at a way to implement this in an version agnostic manner, the general implementation is usable but linking it to go 1.8 can make things difficult.

notzippy:
It has been a long time coming but now a number of dependent libraries only support Go 1.8 as well so I am now merging this. Thanks @zuoRambo

PR #1226, opened 2017-08-19
add map[string]interface{} support for field and binder

runner-mei:
add map[string]interface{} support for field and binder


runner-mei:
I rebase it.

notzippy:
Added via another PR

PR #1234, opened 2017-09-08
Update show.html

jirawat001:


notzippy:
Thanks for the PR, you need to add some context with what you are trying to accomplish. Currently I see you added a Button that is wrapping an input and an anchor. This is not good form, I am closing this but feel free to reopen

PR #1271, opened 2017-11-22
Production patchset

### LINKED ISSUE #1267: 0.18.0's new WebSocket get "interface conversion: interface is nil, not io.Writer" error
In 0.15.0, the old 3rd party websocket works very well, and if upgrade to 0.18.0, the new websocket will get "interface conversion: interface is nil, not io.Writer" error if refresh the page.

I don't know why, but it really happens, even I just by pass all the WS protocol in routes file, it still the same result, and then all the system will get 502 error in all the pages.

Question 1: how it happens and how to fix?
Question 2: could it has a way to use the old websocket or 3rd party websocket???

Thanks.  And now I could only temporary revert to 0.15.0 and it works like a charm...

notzippy:
Fix websocket issue
Modified errors if source file is not found it will not fail
Modified panic if in DevMode a nicely formatted stack is printed.
closes #1267, #1276, #1279

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -55,12 +55,16 @@ func NewErrorFromPanic(err interface{}) *Error {
 	if err != nil {
 		description = fmt.Sprint(err)
 	}
+	lines, readErr := ReadLines(filename)
+	if readErr != nil {
+		utilLog.Error("Unable to read file", "file", filename, "error", readErr)
+	}
 	return &Error{
 		Title:       "Runtime Panic",
 		Path:        filename[len(basePath):],
 		Line:        line,
 		Description: description,
-		SourceLines: MustReadLines(filename),
+		SourceLines: lines,
 		Stack:       stack,
 	}
 }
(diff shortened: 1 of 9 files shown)


errors.go
```diff
@@ -55,12 +55,16 @@ func NewErrorFromPanic(err interface{}) *Error {
 	if err != nil {
 		description = fmt.Sprint(err)
 	}
+	lines, readErr := ReadLines(filename)
+	if readErr != nil {
+		utilLog.Error("Unable to read file", "file", filename, "error", err)
```
runner-mei: utilLog.Error("Unable to read file", "file", filename, "error", err)
to
utilLog.Error("Unable to read file", "file", filename, "error", readErr)

notzippy: Fixed

rami-dabain:
Anyone merge this please? the websocket bug is prohibiting me from updating revel as all of my project is websocket based !

rami-dabain:
sure, ping me if you want a help in review

PR #1277, opened 2017-12-06
fix color is disabled on the windows

runner-mei:
fix color is disabled on the windows


pedromorgan:
Cant quite figure out what this change is about.. is it cli or something ? a muppet needs enlightment

notzippy:
@pedromorgan It adds in a library so that windows terminal output is colored

PR #1290, opened 2018-01-02
Use gofmt -s on all files

mastertinner:
This improves the readability of the project and increases the score on [Go Report Card](https://goreportcard.com/report/github.com/revel/revel).


notzippy:
Definitely a TODO before version 1 release, but until we get there with all the code in a bit of flux with branches I am going to close this until we can do this on a final run

PR #1295, opened 2018-01-12
Fix issue revel/cmd#113

Acidic9:
In some cases (especially during a redirect), module on line 470 returned from `ModuleFromPath()` returned `nil`, causing the application to crash after using `module.Namespace()`.

This issue is addressed here revel/cmd#113.

Commit notzippy@75b7d91 introduced this bug.

### DIFF
--- a/results.go
+++ b/results.go
@@ -469,7 +469,13 @@ func getRedirectURL(item interface{}, args []interface{}) (string, error) {
 		}
 		module := ModuleFromPath(recvType.PkgPath(),true)
 		println("Returned ", module)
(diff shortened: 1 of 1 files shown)


notzippy:
Yes that is what I suspected - however `ModuleFromPath` should never return nil, I believe this is a windows directory matching issue.

notzippy:
@Acidic9 Although your patch treated the sympton it did not fix the root of the issue, I believe #1296 resolves this issue fully. I am going to close this for now, let me know if 1296 fixes this bug as well

PR #1299, opened 2018-01-24
Render a file by a provided filepath

rikvdh:
I was rewriting my Golang web-application where I serve images (listed in my database) from disk, I was repeating myself with the `os.Open()` and though by myself, why isn't this part of the framework. I think this is a good addition to have even more batteries included. :-)

Greetings,

Rik


notzippy:
Looks good, can you also do a PR for the documentation [here](https://github.com/revel/revel.github.io/blob/master/manual/results.md)

PR #1312, opened 2018-02-15
Feature/better required validator

lokhman:
I thought we should allow more types to be validated against `Required` validator. The easiest way to do it is using `reflect`. Now we will be able to validate bit-specific `int`'s, `uint`'s, `float`'s, `struct`'s, etc.

Additionally I standardised default message to be consistent with other validators (having line break at the end).

### DIFF
--- a/validators.go
+++ b/validators.go
@@ -30,17 +30,17 @@ func ValidRequired() Required {
 
 func (r Required) IsSatisfied(obj interface{}) bool {
 	if obj == nil {
-		return true
+		return false
 	}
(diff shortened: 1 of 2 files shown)


notzippy:
Sorry work has been crazzzy, I was playing with this and found [this](https://play.golang.org/p/dyc60wpvREb) scenario where the item is an empty array yet it according to your test it is not empty. Can you update your code so this works as well ?

lokhman:
I have amended the code to match zero-length values and added extra test case. CI build fails due to some other irrelevant issue (probably requires redeployment).

Issue #1323, opened 2018-03-20
gopkg.in/fsnotify.v1: unrecognized import path


gourlaa:
Hi,

When we get revel like this :
` go get github.com/revel/revel`

Since today, we have 
`package gopkg.in/fsnotify.v1: unrecognized import path "gopkg.in/fsnotify.v1" (parse https://gopkg.in/fsnotify.v1?go-get=1: no go-import meta tags ())`

Do you know why?

Thanks,

gourlaa:
Apparently that repo no longer exists :/

nikolawannabe:
https://github.com/fsnotify/fsnotify/issues/118 Author claims to have removed that a long time ago.  Proper url linked in attached commit.

alika:
Related issue https://github.com/go-fsnotify/fsnotify/issues/1. Seems this repo was just created today https://github.com/go-fsnotify. Which also corresponds to when my build that references `gopkg.in/fsnotify` started breaking.

gourlaa:
Ok there change the location of the repo : https://github.com/go-fsnotify/fsnotify/blob/master/README.md

**Change your dependencies from github.com/go-fsnotify/fsnotify to github.com/fsnotify/fsnotify**

Liangzx:
I use "github.com/fsnotify/fsnotify" run well

gourlaa:
I just did a pull request : https://github.com/revel/revel/pull/1324

As with Docker, we can't update anymore our application..

ptman:
May I recommend [dep](https://github.com/golang/dep/) or something else that allows you to use your own copies of dependencies? That way you can also easily patch stuff

notzippy:
Released version 0.19.1 to address this issue - thanks for the quick work guys

HemanthKumarChinta:
Hi , Any idea on the error please ?

   gopkg.in/mgo.V2: cannot find module providing package gopkg.in/mgo.V2: unrecognized import path "gopkg.in": parse https://gopkg.in/?go-get=1: no go-import meta tags ()

PR #1324, opened 2018-03-21
Fix import to point to the fsnotify/fsnotify.v1.

gourlaa:
@https://github.com/go-fsnotify/fsnotify/issues/1
@https://github.com/revel/revel/issues/1323


yatuhashi:
Great! Thanks!

PR #1329, opened 2018-04-05
Updates fsnotify import path due to import path vulnerability

otto-md:
Due to the vulnerability discussed in  go-fsnotify/fsnotify#1, this patch corrects the import path for fsnotify.

same fix as #1325 but for 0.16 branch.


notzippy:
thanks!

PR #1358, opened 2018-08-20
fix template_engine.go's DEBUG bug of command "revel version";

Heinoc:
When I use command "revel version" to check the revel's version, I saw a bug there, and the below is the screenshot.
I have reviewed the code, and find in line 45 of "template_engine.go" have something wrong.
`templateLog.Debug("Registered template engine loaded", key)`
according to the log15, we missed one parameter. and I change it to
`templateLog.Debug("Registered template engine loaded", "name", key)`

Please check is it right, thanks!

image


notzippy:
Thanks @Heinoc

PR #1377, opened 2018-10-02
Correct Revel installation instructions

dbriemann:
Because the Revel command does not include the revel package anymore it's not sufficient to do 
`go get -u github.com/revel/cmd/revel`. I just ran into this problem and thought the Readme might need a little update here.


dbriemann:
Ah ok. I assume the revel command does not automatically update the revel framework? Because I had an older version of revel in my Gopath.

PR #1387, opened 2018-10-08
Session engine 3

notzippy:
Enhancement pack for next release
Added session engine support, and the session cookie engine
breaking change revel.Session was map[string]string now is map[string]interface{}


lucastetreault:
Changing revel.Session from map[string]string to map[string]interface{} is super annoying. Now I'm forced to do type conversions all over my code. It's probably too late to argue against this change given that it has been merged... but if it were up to me I would put this back the way it was and provide convenience methods to marshal/unmarshal into the session.

notzippy:
@lucastetreault  There are methods to unmarshal data see https://godoc.org/github.com/revel/revel/session#Session.GetInto . 
* `Get` fetches the object from the session, if it is an object it will retrieve it as a map[string]interface{}
*  `GetInto` fetches the object from the session into the passed in object
* `GetDefault` fetches the object from the session into the passed in object, or returns the default value if not found

You do not need to manually marshal the data since it is automatically done for you. The only extra convenience method could be something like `GetString()` which could return a string from the session. A PR on this would be welcome.

PR #1401, opened 2018-10-27
Bump Go versions and use '.x' to always get latest minor versions

HaraldNordgren:


HaraldNordgren:
All Windows builds are failing. Not exactly sure why.

notzippy:
Thanks, can you do PR's for revel/cmd, revel/modules as well ?
Windows builds are failing because travis never officially supported windows until recently and right now travis only partially supports windows.

HaraldNordgren:
@notzippy Yup, see https://github.com/revel/cmd/pull/157.

Issue #1424, opened 2019-03-05
DoS vulnerability in Revel framework


SYM01:
Once the [slices parameter](https://revel.github.io/manual/parameters.html#slices) feature is used, the website will be suffering from DoS attack.

When we need to get a slice parameter, we may use the following code snippet in our controller.
```golang
func (c App) DoS1(name []string) revel.Result {
	return c.RenderText(fmt.Sprint(name))
}

func (c App) DoS2() revel.Result {
	var name []string
	c.Params.Bind(&name, "name")
	return c.RenderText(fmt.Sprint(name))
}
```

It looks like everything is OK. However, we can exhaust the server's MEM with only one request.
e.g., when simply visit `http://localhost:9000/dos1?name[1234567890]=1`, the server's CPU and memory usage will soar, and the OOM killer will be triggered eventually.

This vulnerability was caused by the following code:
https://github.com/revel/revel/blob/a3d7a7c23ca885cc5036d3641ede49ce4a14ee2e/binder.go#L210-L277

When the function above is invoked, Revel will calc the `maxIndex` from user's input, and thus the attacker could make the `maxIndex` as large as possible. When `reflect.MakeSlice` is called, Golang will alloc a large memory as the `maxIndex` needed.

A  possible solution for this vuln is to specify the upper bound of the `maxIndex` in the code or config file.

Before the problem is fixed, plz avoid using the slices parameter feature.

notzippy:
Yes, I can see how that could be an issue, but would it not be a better idea to place the check on where `maxIndex` is set (note the check added on the `if index > -1 && maxIndex < maxIndexBound {` below) ?
```
	maxIndexBound := Config.IntDefault("params.max_index", 1024)

	// Factor out the common slice logic (between form values and files).
	processElement := func(key string, vals []string, files []*multipart.FileHeader) {
		if !strings.HasPrefix(key, name+"[") {
			return
		}

		// Extract the index, and the index where a sub-key starts. (e.g. field[0].subkey)
		index := -1
		leftBracket, rightBracket := len(name), strings.Index(key[len(name):], "]")+len(name)
		if rightBracket > leftBracket+1 {
			index, _ = strconv.Atoi(key[leftBracket+1 : rightBracket])
		}
		subKeyIndex := rightBracket + 1

		// Handle the indexed case.
		if index > -1 && maxIndex < maxIndexBound {
			if index > maxIndex {
				maxIndex = index
			}
			sliceValues = append(sliceValues, sliceValue{
[...]

PR #1425, opened 2019-03-05
fix issue #1424

SYM01:


binder.go
```diff
@@ -229,6 +231,10 @@ func bindSlice(params *Params, name string, typ reflect.Type) reflect.Value {
 
 		// Handle the indexed case.
 		if index > -1 {
+			// Just ignore illegal index, fix issue #1424
+			if index > maxIndexBound {
+				return
```
notzippy: Could you a a log statement here like
```suggestion
				binderLog.Error.Println("Invalid parameter index, ignoring parameter","index", maxIndex,"key",key)
				return
```

SYM01:
All right, I will make a new pr with a hotfix branch.

PR #1434, opened 2019-04-02
Fix bug where action is invoked even when Before interceptor returns non-nil value

### LINKED ISSUE #1433: Redirects are ignored in a BEFORE interceptor function
I'd like to be able to redirect to a URL from a `Before` interceptor. This doesn't appear to work:

```go
type BaseController struct {
	*revel.Controller
}

func (c *BaseController) Before() (revel.Result, *BaseController) {
	revel.AppLog.Info("Running before")
	return c.Redirect("http://www.google.com"), c
}
```

I get the log output, but the redirect is ignored. How can I redirect from an interceptor function? 

Minimal example application: https://github.com/dmjones/revel-interceptor-issue

dmjones:
Adjusts the `BeforeAfterFilter` so that it doesn't continue to execute interceptors or actions if the BEFORE function returns a value.

Adds a new test file, `before_after_filter_test.go`, which verifies this behaviour.

Fixes #1433.

### DIFF
--- a/before_after_filter.go
+++ b/before_after_filter.go
@@ -22,6 +22,7 @@ func BeforeAfterFilter(c *Controller, fc []Filter) {
 	}()
 	if resultValue := beforeAfterFilterInvoke(BEFORE, c); resultValue != nil && !resultValue.IsNil() {
 		c.Result = resultValue.Interface().(Result)
+		return
 	}
 	fc[0](c, fc[1:])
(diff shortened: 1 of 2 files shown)


dmjones:
Honestly a bit stuck on what those CI errors are caused by...

notzippy:
TY for the contribution, it was the original intent to do this.

PR #1481, opened 2020-01-30
Feature/domain validator min length

artonge:
I setted min length to 4 because I think a domain must at least look like that 'a.bc'. But maybe it is only 3 'a.b'.

I need that change because I have an error when the domain is an empty string.

### DIFF
--- a/validators.go
+++ b/validators.go
@@ -409,6 +409,11 @@ func (d Domain) IsSatisfied(obj interface{}) bool {
 			return false
 		}
 
+		//can't be less than 4 chars.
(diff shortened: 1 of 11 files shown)


validators.go
```diff
@@ -409,6 +409,11 @@ func (d Domain) IsSatisfied(obj interface{}) bool {
 			return false
 		}
 
+		//can't be less than 4 chars.
+		if l < 4 {
```
notzippy, line 413: Change this to 3 characters. (for `a.b` is valid)

PR #1497, opened 2020-07-02
Update server_adapter_go.go

lujiacn:
Avoid message "INFO  02:07:49    app server_adapter_go.go:499: http: superfluous response.WriteHeader call from github.com/revel/revel.(*GoResponse).WriteStream (server_adapter_go.go:499)   section=system"


notzippy:
Thanks looks good.

PR #1498, opened 2020-07-03
SameSite cookie support

chrishulbert:
Support SameSite cookie parameter

This has become an issue when developing locally, because most browsers now give warnings if you don't have SameSite set, and the warning states they will soon ban cookies that don't have SameSite or Secure set. So this would allow you to set SameSite so it'll work when developing locally over HTTP not HTTPS.

I'm not sure, but maybe it should default to 'strict' for local development so people don't get stung by this when browsers begin to enforce this rule; it seems like it'd be safer to leave this as 'default' for production so as to not break anyone's sites - however it might even better to default it to 'strict' for production for new sites too? Open to the maintainers thoughts on those issues.

Thanks all

### DIFF
--- a/flash.go
+++ b/flash.go
@@ -59,6 +59,7 @@ func FlashFilter(c *Controller, fc []Filter) {
 		Value:    url.QueryEscape(flashValue),
 		HttpOnly: true,
 		Secure:   CookieSecure,
+		SameSite: CookieSameSite,
 		Path:     "/",
 	})
 }
--- a/revel.go
+++ b/revel.go
@@ -6,6 +6,7 @@ package revel
 
 import (
 	"go/build"
+	"net/http"
 	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


chrishulbert:
Documentation to match: https://github.com/revel/revel.github.io/pull/196

ptman:
I believe this is needed and an improvement. Apparently extra cookies can set their samesite property freely. What about i18n cookie?

chrishulbert:
Thanks @ptman for getting back about this.
I had a look for i18n cookies, and found that it reads the cookie in i18n.go > hasLocaleCookie, however I could not find where that cookie is set anywhere? Is this intended for the framework consumer to set that cookie or am i missing something?
Thanks again

ptman:
Oh, my bad. Seems like revel doesn't set the locale cookie, only reads it if it's set.

ptman:
what's up with the failing tests?

notzippy:
> what's up with the failing tests?

There is an issue with memcache an external project not finding the key occasionally during the tests.

PR #1514, opened 2020-10-02
Fix misspellings and some lint errors

ptman:
This shouldn't cause any change in behaviour


brendensoares:
Looks like a lot of good polishing here. Any comments @notzippy?

notzippy (approved):

PR #1524, opened 2021-01-29
More complicated linting

ptman:
I expect there will be questions and I'll have to reverse some of the changes.

### DIFF
--- a/binder_test.go
+++ b/binder_test.go
@@ -23,7 +23,7 @@ type A struct {
 	ID   int
 	Name string
 	B    B
-	//nolint:unused
+	//nolint:unused,structcheck
 	private int
 }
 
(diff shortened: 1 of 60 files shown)


notzippy:
This seems to have errors with the results class
```
./results.go:146:25: Errorf format %w has arg err of wrong type interface{}
./results.go:194:10: Errorf format %w has arg rerr of wrong type interface{}
./results.go:211:10: Errorf format %w has arg rerr of wrong type interface{}
```

ptman:
Thanks for the review. I was misled by err := recover(), but recover returns interface{}, not error.

notzippy:
@brendensoares Thoughts on this?

brendensoares:
This looks good, but we'll defer to next release pending testing/verification.

PR #1546, opened 2021-11-03
Correction garyburd/redigo import

gourlaa:
The package github.com/garyburd/redigo doesn't exist anymore.
Just replaced the import by the new package : https://github.com/gomodule/redigo/


brendensoares:
Is this a 1:1 compatible interface?

PR #1555, opened 2022-06-05
Added X-HTTP-Method-Override validation

movitz-s:
This MR will make sure that the value passed in `X-HTTP-Method-Override` is a valid HTTP token as required in the RFC. The validation resolves a security issue that can occur when Revel is used behind reverse proxies. Because of how the route matching is done, one could make a request to `/site` with the header `X-HTTP-Method-Override: POST/admin` and it would be routed to `/admin/site`. This could be exploited if the security setup relied on a proxy blocking certain path or if the proxy added a trusted prefix to every request.  

I don't know if this is the best way of resolving the issue (or if it is even fully resolved). If you want to implement it another way please do, I just wanted to bring it to attention.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/mattn/go-isatty v0.0.14 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e // indirect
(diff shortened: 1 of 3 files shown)


movitz-s:
KilledKenny's fix looks better and also has tests, so I'll close this PR. Shout-out to those who found this vuln at the Security Fest CTF in Gothenburg last week (I personally didn't).

PR #1556, opened 2022-06-05
Added verification of X-HTTP-Method-Override before use

KilledKenny:
Verify that the method in X-HTTP-Method-Override is a valid method
before use using it. Validation is performed using the same logic
as net/http dose. This prevents a routing confusion vulnerability
that allowed an attacker to control the entire internal revel routing
path (used to find a controller) via the override header. This issue can
be problematic in instances where authentication is based on the path
for example in a revel.Filter or if the revel app is behind a reverse
proxy.

### DIFF
--- a/router.go
+++ b/router.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/revel/pathtree"
 	"github.com/revel/revel/logger"
+	"golang.org/x/net/http/httpguts"
 )
 
 const (
@@ -160,6 +161,16 @@ func treePath(method, path string) string {
 	return "/" + method + path
 }
 
+//validMethod is copied from net/http/request.go
+func validMethod(method string) bool {
+	return len(method) > 0 && strings.IndexFunc(method, isNotToken) == -1
+}
+
+//isNotToken is copied from net/http/http.go
+func isNotToken(r rune) bool {
+	return !httpguts.IsTokenRune(r)
+}
+
 type Router struct {
 	Routes []*Route
 	Tree   *pathtree.Node
@@ -169,7 +180,7 @@ type Router struct {
 
 func (router *Router) Route(req *Request) (routeMatch *RouteMatch) {
 	// Override method if set in header
-	if method := req.GetHttpHeader("X-HTTP-Method-Override"); method != "" && req.Method == "POST" {
+	if method := req.GetHttpHeader("X-HTTP-Method-Override"); method != "" && req.Method == "POST" && validMethod(method) {
 		req.Method = method
 	}
 
--- a/router.go
+++ b/router.go
@@ -162,13 +162,13 @@ func treePath(method, path string) string {
 }
 
 //validMethod is copied from net/http/request.go
-//its used to validate the X-HTTP-Method-Override so that it only may contain valid characters for a http method in a identical way to how net/http validates method.
+// used to validate the X-HTTP-Method-Override so that only valid HTTP method characters are allowed (in the same way as in net/http/request.go).
 func validMethod(method string) bool {
 	return len(method) > 0 && strings.IndexFunc(method, isNotToken) == -1
 }
 
 //isNotToken is copied from net/http/http.go
-//its is used by validMethod to validate X-HTTP-Method-Override
+// used by validMethod to detect unacceptable characters in X-HTTP-Method-Override
 func isNotToken(r rune) bool {
 	return !httpguts.IsTokenRune(r)
 }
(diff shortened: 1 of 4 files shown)


KilledKenny:
this fix is almost identical to  #1555. The functional difference is that my fix ignores the header when its invalid while movitz-s fix will return nil if the header is invalid.

notzippy:
This looks good to me - @brendensoares thoughts?

router.go
```diff
@@ -169,7 +180,7 @@ type Router struct {
 
 func (router *Router) Route(req *Request) (routeMatch *RouteMatch) {
 	// Override method if set in header
-	if method := req.GetHttpHeader("X-HTTP-Method-Override"); method != "" && req.Method == "POST" {
+	if method := req.GetHttpHeader("X-HTTP-Method-Override"); method != "" && req.Method == "POST" && validMethod(method) {
```
brendensoares, line 186: It would be valuable to add a comment here for the context of why this line is necessary. Context for our future coders and our future selves.

router.go
```diff
@@ -160,6 +161,16 @@ func treePath(method, path string) string {
 	return "/" + method + path
 }
 
+//validMethod is copied from net/http/request.go
```
brendensoares, line 164: Can we add more doc/context here? I realize the original method you cited doesn't have a doc block, but it does at least have an internal comment block that shows what it's testing for.

See https://github.com/golang/go/blob/a5d61be040ed20b5774bff1b6b578c6d393ab332/src/net/http/request.go#L817

router.go
```diff
@@ -160,6 +161,16 @@ func treePath(method, path string) string {
...
+func validMethod(method string) bool {
+	return len(method) > 0 && strings.IndexFunc(method, isNotToken) == -1
+}
+
+//isNotToken is copied from net/http/http.go
+func isNotToken(r rune) bool {
```
brendensoares, line 172: Same as above. At least some comment like "tests for HTTP methods."

brendensoares (changes requested):

brendensoares:
I appreciate the unit tests you included, but could you also provide steps to reproduce the vulnerability so we can verify this correctly addresses it?

Thank you for contributing to Revel and OSS!

router.go
KilledKenny: Improved the comment

router.go
KilledKenny: Improved the comment

router.go
KilledKenny: added referents to validMethod in the comment

router.go
```diff
@@ -162,11 +162,13 @@ func treePath(method, path string) string {
 }
 
 //validMethod is copied from net/http/request.go
+//its used to validate the X-HTTP-Method-Override so that it only may contain valid characters for a http method in a identical way to how net/http validates method.
 func validMethod(method string) bool {
 	return len(method) > 0 && strings.IndexFunc(method, isNotToken) == -1
 }
 
 //isNotToken is copied from net/http/http.go
+//its is used by validMethod to validate X-HTTP-Method-Override
```
brendensoares: Can we use:

> // used by validMethod to detect unacceptable characters in X-HTTP-Method-Override

router.go
```diff
@@ -162,11 +162,13 @@ func treePath(method, path string) string {
 }
 
 //validMethod is copied from net/http/request.go
+//its used to validate the X-HTTP-Method-Override so that it only may contain valid characters for a http method in a identical way to how net/http validates method.
```
brendensoares: Could we use the following:

> // used to validate the X-HTTP-Method-Override so that only valid HTTP method characters are allowed (in the same way as in net/http/request.go).

router.go
brendensoares: I mean we should note that this is required to negate the vulnerability. For example, you described to me privately this way:

> We found that it was possible to control the http req path via the header X-HTTP-Method-Override. This issue can be problematic in instances where authentication is based on the path for example in a revel.Filter or if the revel app is behind a reverse proxy.

brendensoares (changes requested):

brendensoares:
Added new code review comments.

Also, I got your proof of concept code to test for this vulnerability in gitter. Thank you.

router.go
KilledKenny: Used your comments

router.go
KilledKenny: Used your comments

router.go
KilledKenny: added more meat to the comment

